PR #18, opened 2021-08-10
Decouple tasks from code

mcharleb:
The tests should be dynamic based on the tasks.pbtxt file not hardcoded as they currently are in the app. Adding a new test that uses existing datasets requires code changes in about 3 places.


bellycat77:
Is it expected behavior after applying the current patch?
Screenshot 2021-08-16 at 22 30 32

mcharleb:
Yes, this is expected because the models are not known until tasks_vt.pbtxt
is loaded via http or via the sdcard

bellycat77 (approved):

PR #70, opened 2021-09-16
Use protobuf-3.17.3

bellycat77:


mcharleb:
@bellycat77 How is it that CI passes if in the issues comments the build fails on an actual phone?

bellycat77:
@mcharleb 
> @bellycat77 How is it that CI passes if in the issues comments the build fails on an actual phone?

The PR was tested on my phone Xiaomi Redmi 9, also it was successfully tested by CI on Google Pixel 5e actual phone. So it is not a trivial fault.

Issue #55, opened 2021-09-08
Potential problem with cached files from mobile_models repo


bellycat77:
We have a potential problem with cached files from the https://github.com/mlcommons/mobile_models repo

We have 2 files with equivalent names but different links:
https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_float.tflite
https://github.com/mlcommons/mobile_models/blob/main/v1_1/tflite/mobilenet_edgetpu_224_1.0_float.tflite
The application "looks" at the filename when it checks if was the file cached or not. So if we will not remove the old application version before installing the new one, then cached files will be kept. And if the new version of the application requires a new version of the file, then it will no be uploaded, because a file with the same name is already in cache.

relja128:
That's annoying. Can you cache by url instead? It will conveniently evict the existing files.

d-uzlov:
> Can you cache by url instead?

Yes, this would solve this issue.

>It will conveniently evict the existing files.

Using full URL in cache will not help evicting existing files. Our current cache system (both in android and in flutter) only downloads missing files. If we no longer use the old files, they just lie in the file system until the app is uninstalled, or user manually clears the cache.

Keeping the cache clean is a different, more complicated issue.

Except for the simplest solution of using full URL in path, which doesn't help with eviction of old files, there are several solutions, which can fix the eviction issue:
1. Use full URL for paths, automatically purge cache when app is updated.
**Pros**: simple.
**Cons**: we will download all data from scratch after each update.
2. Make a more sophisticated caching system, and try to keep track of the files, to delete files that were needed previously but no longer needed in the new version.
**Cons**: complicates code, possible overhead for tracking files.
3. Use some external library to help with caching.
**Pros**: less complicated code.
**Cons**: added dependency, possibly even more overhead compared to [solution 2].

Also, our apps can use custom tasks.pbtxt files, and it's not clear what we should do with the cache for these custom files, if we plan on evicting old files.

relja128:
Let's go with cache by url. Evict when app is updated AND files are more
than 9 months old. (Both conditions have to be true). So after the app
version changes, delete files older than 9 months.

anhappdev:
How about using a file checksum like MD5?
The config file should have a new field storing the checksum of the ML model in the URL. The app compares the checksum of the cached file with the checksum in the config file and download the new file if they are different. The filename can still be the same to evict the existing files.

[...]

PR #72, opened 2021-09-21
Fix issue #55 with downloaded files caching

bellycat77:


android/java/org/mlperf/inference/MLPerfTasks.java
```diff
@@ -115,25 +119,39 @@ public static boolean isZipFile(String path) {
...
+    boolean isURL = path.startsWith("http://") || path.startsWith("https://");
+    if (isURL || isZipFile(filename)) {
       if (isZipFile(filename)) {
         filename = filename.substring(0, filename.length() - ZIP.length());
       }
-      String localDir =
-          MLCtx.getInstance().getContext().getExternalFilesDir("cache").getAbsolutePath();
-      return localDir + "/cache/" + filename;
+
+      if (isURL) {
```
mcharleb, line 137: Shouldn't this be "else if"

android/java/org/mlperf/inference/MLPerfTasks.java
```diff
@@ -115,25 +119,39 @@ public static boolean isZipFile(String path) {
...
+        String longFileName;
+        try{
+          longFileName = new URL(path).getPath();
+        }
+        catch(Exception e) {
+          longFileName = filename;
```
mcharleb, line 143: What failure condition of URL(path).getPath() will this fix?

mcharleb: Does File(path).getName() of a URL produce the expected result?

android/java/org/mlperf/inference/activities/CalculatingActivity.java
```diff
@@ -1294,6 +1347,7 @@ protected boolean extractFileFail(String src) {
           if (!unZip(tmpFile, dest)) {
             return true;
```
mcharleb, line 1348: Why not delete tmpFile before return if the unzip failed?

mcharleb (changes requested):

android/java/org/mlperf/inference/MLPerfTasks.java
bellycat77: No, it is different "if".

android/java/org/mlperf/inference/MLPerfTasks.java
bellycat77: First "if" drops ".zip" extension if presented, second "if" keep "path" from URL.
The resulting name of the file will be something /{cache directory path}/{file name} for locally stored file
and /{cache directory path}/{path from url}/{file name} for URL. In both cases ".zip" extension will be dropped if presented. So if downloaded files have the same name, but different URLs - they will be stored in different folders in the cache

android/java/org/mlperf/inference/activities/CalculatingActivity.java
bellycat77: Fixed

android/java/org/mlperf/inference/activities/CalculatingActivity.java
```diff
@@ -248,6 +252,62 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+      isVersionChanged = !storedVersion.equals(currentAppVersion);
+      Log.i(TAG, "Version is changed: " + isVersionChanged);
+      long diff = new Date().getTime() - cacheDir.lastModified();
+      Log.i(TAG, "Stored version is: " + storedVersion);
+
+      if (diff > Long.valueOf(AppConstants.MAX_FILE_AGE_IN_DAYS) * 24 * 60 * 60 * 1000
+          && isVersionChanged) {
+        try {
+          FileUtils.forceDelete(cacheDir);
```
anhappdev, line 292: The app freezes for 30 seconds when this is called.

android/java/org/mlperf/inference/MLPerfTasks.java
anhappdev: The code seems hard to understand for a quick scan. BTW, the path of some folders still have a `.zip` suffix like this one: `/sdcard/mlperf_results/mlperf/cache/mlcommons/mobile_models/raw/main/v0_7/datasets/imagenet.zip`

android/java/org/mlperf/inference/MLPerfTasks.java
```diff
@@ -32,6 +33,9 @@
 public final class MLPerfTasks {
   private static final String TAG = "MLPerfTasks";
   private static final String ZIP = ".zip";
+  private static final String LOCAL_FOLDER = "/sdcard/mlperf_results/mlperf/";
+  private static final String LOCAL_CACHE_FOLDER = LOCAL_FOLDER + "cache/";
```
anhappdev, line 37: LOCAL_CACHE_FOLDER will be `/sdcard/mlperf_results/mlperf/cache`. The path is a bit unusual. Should LOCAL_CACHE_FOLDER better be `/sdcard/mlperf_cache`?

PR #75, opened 2021-09-22
Remove unused tflite_jni_binary from cpp backend BUILD

stegben:


mcharleb (approved):
Seems to pass CI without it.

PR #83, opened 2021-09-28
Fix code formatting

bellycat77:


mcharleb:
@bellycat77  The style in this PR and in https://github.com/mlcommons/mobile_app_open/pull/81/files are not the same. Only one of these can be correct.

mobile_back_qti/cpp/backend_qti/allocator.cc
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 #include "allocator.h"
 
 // Map of allocation size to Chunk allocator instance
-std::map<size_t, ChunkAllocator> &getAllocator() {
+std::map<size_t, ChunkAllocator>& getAllocator() {
```
mcharleb, line 19: Which style is correct, this or https://github.com/mlcommons/mobile_app_open/pull/81/files ?

mcharleb (changes requested):

mobile_back_qti/cpp/backend_qti/allocator.cc
bellycat77: This one is correct. It was checked by fixed CI. The mentioned file was not changed by PR #81, so it was not checked by CI in the case of PR #81.

mcharleb (approved):

Issue #92, opened 2021-10-06
Open source the Flutter app


relja128:
Let's discuss, and if everyone is on board, we can proceed with developing in mobile_app_open.

relja128:
1) Flutter repo becomes a folder in mobile_app_open: flutter
2) Merge workspace files
3) Squash the history (copy the files over)
---
Later we will re-organize the back ends

mcharleb:
Code was merged

PR #103, opened 2021-11-01
Merge android and flutter bazel configs

d-uzlov:
Flutter app was manually tested with these changes, Windows and iOS versions work.

Part of https://github.com/mlcommons/mobile_app_open/issues/92
Continuation of https://github.com/mlcommons/mobile_app_open/pull/102
Closes https://github.com/mlcommons/mobile_app_open/issues/107


.bazelrc
```diff
@@ -1,28 +1,72 @@
...
+build:windows --cxxopt=/std:c++14
+build:windows --host_cxxopt=/std:c++14
+
+# Android configs
 build:android --crosstool_top=//external:android/crosstool
 build:android --host_crosstool_top=@bazel_tools//tools/cpp:toolchain
-
 build:android_arm64 --config=android
 build:android_arm64 --cpu=arm64-v8a
 build:android_arm64 --fat_apk_cpu=arm64-v8a
```
mcharleb, line 33: Do we need a fat APK anymore if it only contains one arch?

mcharleb (approved):

.bazelrc
d-uzlov: According to Tensorflow config we do need both flags:
https://github.com/tensorflow/tensorflow/blob/master/.bazelrc#L158-L160
```bazel
# Android configs. Bazel needs to have --cpu and --fat_apk_cpu both set to the
# target CPU to build transient dependencies correctly. See
# https://docs.bazel.build/versions/master/user-manual.html#flag--fat_apk_cpu
```

mcharleb (changes requested):
README.md should also be updated with new directory structure

mcharleb (approved):

Issue #113, opened 2021-11-02
Pull cpp folder out of android and flutter directories


d-uzlov:
We have 2 mostly duplicated `cpp` folders: [android/cpp](https://github.com/mlcommons/mobile_app_open/tree/347d490c1ef90cf62a37ec097ba8d846a1b8ead6/android/cpp) and [flutter/cpp](https://github.com/mlcommons/mobile_app_open/tree/347d490c1ef90cf62a37ec097ba8d846a1b8ead6/flutter/cpp).
Folders were last synced several weeks ago, so there are few differences.

We need to pull `cpp` directory to to the top level near `android` and `flutter` directories, merging its contents.

relja128:
Leave as-is. We will deprecate Android in ~6 months.

relja128:
Having them in the same location helps with build - backends would pull headers in from only one spot. Let's do this.

PR #114, opened 2021-11-02
Pull cpp folder out of android and flutter directories

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/113


d-uzlov:
Bazel configs in `/cpp` folder have references to `/android/commonlibs`. Not sure if I should move `commonlibs` folder to the root directory, or move it to the `/cpp` directory (to avoid cluttering the root directory), or leave it as it is.

anhappdev:
@d-uzlov Can you merge master into this branch? It will be easier to test then. Thanks.

d-uzlov:
@anhappdev there is nothing to be merged. I rebased the branch onto fresh commit from master before marking this PR as ready for review.

anhappdev:
I got an error related to wrong path in Makefile, which is fixed in `master`. But I see you just pushed a commit to fix it.

anhappdev (changes requested):
Path in file `mobile_app_open/flutter/ios/Runner/Runner-Bridging-Header.h` need to be updated to `"../../../cpp/flutter/main.h"`

anhappdev:
> Bazel configs in `/cpp` folder have references to `/android/commonlibs`. Not sure if I should move `commonlibs` folder to the root directory, or move it to the `/cpp` directory (to avoid cluttering the root directory), or leave it as it is.

There is only a BUILD file there, but seems to include both Android-specific config and shared config. I would split the file to decouple it (maybe in another PR).

anhappdev:
The folder `mobile_app_open/cpp/flutter` is only for Flutter app relevant. Should we move it into `mobile_app_open/flutter` (in another PR)?

anhappdev (approved):

d-uzlov:
> The folder `mobile_app_open/cpp/flutter` is only for Flutter app relevant. Should we move it into `mobile_app_open/flutter` (in another PR)?

This folder contains flutter-to-native bindings, similar to JNI code for Android. Expect it just builds a native C-interface dynamic library that is needed for Flutter.
In Android there seems to be a convention to put JNI files into source tree. I can't find any similar conventions for Flutter, but considering we just get a `.framework`/`.so`/`.dll` out of this, it makes sense to keep it in _some_ `cpp` directory.

I moved it into _the_ `/cpp` directory to keep all sources together.
Alternatively we would probably have `/cpp` for general native code, and then a `/flutter/cpp/flutter` folder with few files that are tightly coupled to the `/cpp`. I thought this would be harder to navigate. Though, we do already have separate place for the Android JNI files.

I don't have a strong opinion on where `cpp/flutter` must be located. If you think we should keep it inside `/flutter`, then we probably should.
Though, `/flutter/cpp/flutter` path seems a bit awkward to me, so maybe we should rename it?

@mcharleb, what's your opinion about this?

d-uzlov:
>There is only a BUILD file there, but seems to include both Android-specific config and shared config. I would split the file to decouple it (maybe in another PR).

Probably in another PR. This PR is already a big refactoring, let's not make it even more complex.

d-uzlov:
I moved cpp folder into android/cpp, as we decided on meeting.
I moved cpp/flutter to flutter/cpp/flutter, because it would make little sense to place it in android/cpp/flutter.

mcharleb (approved):
LGTM

anhappdev (approved):

Issue #116, opened 2021-11-03
Pull tflite backed out of flutter folder


d-uzlov:
Flutter app is using its own copy of the tflite backend.
We need to synchronize it with the [/mobile_back_tflite](https://github.com/mlcommons/mobile_app_open/tree/347d490c1ef90cf62a37ec097ba8d846a1b8ead6/mobile_back_tflite) folder, remove [/flutter/cpp/backend_tflite](https://github.com/mlcommons/mobile_app_open/tree/347d490c1ef90cf62a37ec097ba8d846a1b8ead6/flutter/cpp/backend_tflite) folder, and point flutter to the global tflite backend.

Parent issue: https://github.com/mlcommons/mobile_app_open/issues/113

PR #117, opened 2021-11-03
Pull tflite backed out of flutter folder

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/116


mobile_back_tflite/cpp/backend_tflite/BUILD
```diff
@@ -29,6 +29,7 @@ cc_library(
     srcs = ["tflite_c.cc"],
     hdrs = [
         "tflite_settings.h",
+        "tflite_settings_apple.h",
```
mcharleb, line 32: Should this be conditional on ios build?

mcharleb (approved):

mobile_back_tflite/cpp/backend_tflite/BUILD
d-uzlov: We could make this conditional, but in my opinion this would make build config more complicated without any benefits.

d-uzlov: @mcharleb are you OK with current solution or should I change it?

anhappdev:
@d-uzlov Please test if the Android app build still works. I got some error while building the Android app with this PR.

d-uzlov:
@anhappdev do you mean the native Android app or the Android version of the Flutter app?

anhappdev:
> @anhappdev do you mean the native Android app or the Android version of the Flutter app?

The native Android app.

d-uzlov:
Command `bazel build --config android_arm64 //android/java/org/mlperf/inference:mlperf_app` doesn't produce any errors for me locally.
I can't check building with docker locally, but our Jenkins CI doesn't seem to have any issues building the app using the docker container, all jobs for this PR are green.

Could you recheck if these build errors you see are caused by this PR and not by some of your local changes in repo or in environment?

anhappdev:
I used the command `bazel build -c opt //android/java/org/mlperf/inference:mlperf_app` from the [Android README.md](https://github.com/mlcommons/mobile_app_open/tree/master/android#option-1---building-manually) and it failed.

The command `bazel build --config android_arm64 //android/java/org/mlperf/inference:mlperf_app` did work.

anhappdev:
@d-uzlov I can confirm building locally with docker also works. Since you are already on this PR, please update the command in [the Android README.md](https://github.com/mlcommons/mobile_app_open/blob/master/android/README.md#option-1---building-manually). Thanks.

d-uzlov:
I tested this command on different commits in master. Seems like it stopped working after previous commit, which merged #103.
Before that merge default parameters in `.bazelrc` contained settings for android build, so building without specifying anything additional worked.
After that merge android settings are only defined when using appropriate `--config` value.

> Since you are already on this PR, please update the command in the Android README.md

Yep, this is what I was going to suggest.

anhappdev (approved):

anhappdev (changes requested):
I think path need to be updated in `flutter/Makefile`
```
_bazel_ios_fw := ../bazel-bin/flutter/cpp/flutter/ios_backend_fw_static_archive-root/ios_backend_fw_static.framework
```

mobile_back_tflite/cpp/backend_tflite/BUILD
mcharleb: I'm OK with it

anhappdev (approved):

PR #122, opened 2021-11-03
Changes needed when using corporate web proxy

mcharleb:
Added a top level target to build flutter android app using docker image:

make flutter_android

### DIFF
--- a/flutter/Makefile
+++ b/flutter/Makefile
@@ -20,6 +20,13 @@ else
 all: ios
 endif
 
+include ../mobile_back_qti/make/qti_backend.mk
+ifeq (${USE_PROXY_WORKAROUND},1)
+  export PROXY_WORKAROUND=--host_jvm_args -Djavax.net.ssl.trustStore=/etc/ssl/certs/java/cacerts
+else
+  export PROXY_WORKAROUND=
+endif
+
 .PHONY: ios
 ios: cpp-ios prepare-flutter update-splash-screen
 
@@ -143,7 +150,7 @@ backend-bridge-windows:
 
 .PHONY: backend-bridge-android
 backend-bridge-android:
-	bazel build ${_bazel_links_arg} --config=android_arm64 -c opt //flutter/cpp/flutter:libbackendbridge.so
+	bazel ${PROXY_WORKAROUND} build --config=android_arm64 -c opt //flutter/cpp/flutter:libbackendbridge.so
 	cd .. && chmod +w ${BAZEL_LINKS_DIR}bin/flutter/cpp/flutter/libbackendbridge.so
 	mkdir -p android/app/src/main/jniLibs/arm64-v8a
 	rm -f android/flutter/app/src/main/jniLibs/arm64-v8a/libbackendbridge.so
@@ -177,7 +184,7 @@ backends/tflite-windows:
 
 .PHONY: backends/tflite-android
 backends/tflite-android:
-	bazel build ${_bazel_links_arg} --config=android_arm64 -c opt //mobile_back_tflite:tflitebackend
+	bazel ${PROXY_WORKAROUND} build --config=android_arm64 -c opt //flutter/cpp/backend_tflite:libtflitebackend.so
 	cd .. && chmod +w ${BAZEL_LINKS_DIR}bin/mobile_back_tflite/cpp/backend_tflite/libtflitebackend.so
 	mkdir -p android/app/src/main/jniLibs/arm64-v8a
 	rm -f android/app/src/main/jniLibs/arm64-v8a/libtflitebackend.so
--- a/flutter/Makefile
+++ b/flutter/Makefile
@@ -35,7 +35,7 @@ _bazel_links_arg=--symlink_prefix ${BAZEL_LINKS_DIR} --experimental_no_product_n
 
 .PHONY: cpp-ios
 cpp-ios:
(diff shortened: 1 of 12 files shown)


flutter/Makefile
```diff
@@ -143,7 +150,7 @@ backends/tflite-windows:
 
 .PHONY: backends/tflite-android
 backends/tflite-android:
-	bazel build --config=android_arm64 -c opt //flutter/cpp/backend_tflite:libtflitebackend.so
+	bazel ${PROXY_WORKAROUND} --config=android_arm64 -c opt //flutter/cpp/backend_tflite:libtflitebackend.so
```
d-uzlov, line 153: Seems like you accidentally deleted `build` command from here

d-uzlov (changes requested):

mcharleb:
@d-uzlov can you please have a look at this and see why the flutter build step is failing when running "make flutter_app"

d-uzlov:
I'm not sure which exact error you are getting.
Few notes:
* If the output folder doesn't exist, the make script fails.
* I got some strange error:
```bash
cd flutter && dart pub get
Resolving dependencies... (1.2s)
Because mlcommons_ios_app depends on integration_test from sdk which doesn't exist (the Flutter SDK is not available), version solving
  failed.

Flutter users should run `flutter pub get` instead of `dart pub get`.
```
This is strange because command `(cd flutter && dart pub get)` worked for me without any issues, when I run it manually.
I was able to fix it by replacing the `dart pub get` with suggested `flutter pub get`.
I suspect that you could get an error for missing `dart` command, since, as far as I understand, you don't have flutter installed locally.
* Speaking of installing flutter: why do we download repos for flutter and protoc?
It looks kinda strange. Are we going to fetch all dependencies manually in make scripts? I thought we should bake the dependencies into our docker image.
I suspect that issues with the `dart pub get` command I got may be related to this manual flutter download. Master branch in the flutter repo is not tested to work with our app. And it may simply be unstable.

In the end I got an error `The SDK directory is not writable (/opt/android/sdk)`, both using the make script and manual build, and this error seems to be related to my local environment. The build would probably succeed if I fix my Android SDK installation (or build inside the docker image, where the SDK is installed correctly).

P.S. May I suggest you my view of how to organize make targets and make scripts to make it more convenient?

mcharleb:
@d-uzlov the flutter pub get or dart pub get (same thing basically) works for you because you are not behind a Qualcomm web proxy and firewall. The download does not work inside a docker container even if I try to map the certificate from the host into the container.

I cannot run protoc (because it depends on the dart/flutter pub get) in the docker container nor can I run flutter for the same reason. I cannot apt install protoc or flutter on the host so that is why they are downloaded to output dir and used from there in the build host.

mcharleb:
@d-uzlov  wrote: "I was able to fix it by replacing the dart pub get with suggested flutter pub get.
I suspect that you could get an error for missing dart command, since, as far as I understand, you don't have flutter installed locally."

Both flutter and dart are available in the path. From flutter Makefile:

.PHONY: protobuf
protobuf:
        dart pub get

I was just replicating the steps from Flutter/Makefile

flutter/make/android.mk is to support host machine based build
flutter/make/docker.mk support docker based build

mcharleb:
@d-uzlov flutter also expects (gradle) to be able to modify the Android SDK installation which it cannot do if it is part of the docker image because or write permission and that the changes would not be persistent between runs. Seems like for flutter build, the android SDK needs to be installed as the user running the build. In that case it should likely be in the output dir, or else build need to be run as root (not a good idea)

d-uzlov:
@mcharleb, oh, I see.

`dart pub get` is actually not required every time. It is part of the `protobuf` make target only because it is required for a fresh repo. If you were able to build the container somewhere without internet access restrictions, you could theoretically bake this into the container and just skip running it when building.

`flutter build`, however, is, obviously, required every time, and I'm not sure if it can run without internet access.
I checked right now, and I was able to run both protobuf generation and `flutter build apk` without internet access, but I do have everything required predownloaded, because it's not the first time I run these commands on my PC. I believe flutter downloads some stuff when it first runs.

>and then the changes would not be persistent between runs

Doesn't seem like an issue to me.
Well, if it tries to download and install a bunch more stuff, then this is an issue, but in this case we should just download and bake it into the container manually. Else, if it just modifies configs, I don't see any real disadvantages.

d-uzlov:
>I cannot apt install protoc or flutter on the host so that is why they are downloaded to output dir and used from there in the build host

Are there issues with installing into the container?

mcharleb:
@d-uzlov I wasn't able to run flutter in the container so protoc also would not run because it required dart. I resolved the issue about running in the container and can start to refactor the container to have protoc and flutter again.

There are issues in the code using ANDROID_SDK_HOME vas ANDROID_SDK_ROOT and ANDROID_NDK_HOME instead of ANDROID_NDK_ROOT. The use of ANDROD_HOME also seems incorrect (under android top level tree)

mcharleb:
@d-uzlov  wrote "If you were able to build the container somewhere without internet access restrictions, you could theoretically bake this into the container and just skip running it when building..."

You can't bake the pub get into the creation of the docker image because it depends on the pubspec.yaml file and other files in the mobile_app_open flutter directory.

It only seems possible to run dart/flutter pub get from inside a running docker container with that volume/dir mounted.

anhappdev:
I think this PR is no longer updated. I will close it for now. Feel free to reopen if it is still needed.

Issue #125, opened 2021-11-09
Allow same binary to be used for submission and press release


relja128:
We need a unified binary (we already can make this, but we should tighten the rules to disallow per-vendor front end modifications).
The same binary should be able to generate submission and performance results. We will need to fetch config data from github, and then failing that, fetch from SD card.

[For closed submissions only for now]

anhappdev:
> We will need to fetch config data from github, and then failing that, fetch from SD card.

1. With config data you mean the [tasks_v2.pbtxt](https://github.com/mlcommons/mobile_models/blob/main/v1_0/assets/tasks_v2.pbtxt), right? 
2. Should the path to the config file in SD card just be hard-coded to something like `/sdcard/mlperf_datasets/custom_tasks.pbtxt`?
3. Should the change be made for Android or Flutter or both app?

relja128:
1) With config data you mean the tasks_v2.pbtxt, right?
Yes, but if you need additional config data elsewhere, that's ok.

2) Should the path to the config file in SD card just be hard-coded to something like /sdcard/mlperf_datasets/custom_tasks.pbtxt?
No, see below.

3) Should the change be made for Android or Flutter or both app?
Flutter app.

----

What we want to achieve:
1) The press can run the app as-is, no switches needed
2) Same binary can be validated and then released
3) MLCommons developers have an easy way of running the right dataset on the unified binary

We will continue discussion.

One idea: 
By default, the app will use files on the SD card if found, else it will download to cache directory, and use the cache.
There will be a switch to force SD-card-only or download-only behavior.

anhappdev:
I think we should add a file picker to the setting screen.
In the setting screen, the user can set the paths to different datasets and models.
On the first start, the paths will be populated using the default paths from the default `tasks.pbtxt` file. After that, the user can change it (using a file picker for a local path or just enter text for a URL).
This would be a clean approach without too many custom logics for different user groups.

relja128:
We discussed having an edit box with paths in settings, but decided it was unwieldly because of the number and length of paths.

For now, implement the following:
* By default, the app will use files on the SD card if found, else it will download to cache directory, and use the cache.

We will discuss and design the override switches in settings in a future meeting.

Mostelk:
[...]

PR #133, opened 2021-11-15
Add option to build all backends in the same binary

### LINKED ISSUE #135: Option to build app with multiple backends
Part of #125

anhappdev:
Closes #135 
Part of #125

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -88,8 +88,10 @@ make WITH_QTI=1 app
 ```
 or to enable multiple backends:
 ```
-make WITH_QTI=1 WITH_SAMSUNG=1 WITH_PIXEL=1 app
+make WITH_QTI=1 WITH_MEDIATEK=1 app
 ```
+Note: Follow the instruction of the backend vendor (in `mobile_back_*` folders at root level) to build for that backend.
+Some backends may contain proprietary code. In that case, please contact MLCommons to get help.
 
 # Install the app with the command:
 adb install -r build/mlperf_app.apk
--- a/android/README.md
+++ b/android/README.md
@@ -86,9 +86,9 @@ or to enable the QTI backend:
 ```
 make WITH_QTI=1 app
 ```
-or to enable both the MediaTek and QTI backends:
+or to enable multiple backends:
 ```
-make WITH_QTI=1 WITH_MEDIATEK=1 app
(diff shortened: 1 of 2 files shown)


mcharleb:
@anhappdev  Is there an issue associated with this PR? We typically try to create issues for each PR, and then review the issue list vs reviewing the PRs.

android/README.md
```diff
@@ -86,9 +86,9 @@ or to enable the QTI backend:
 ```
 make WITH_QTI=1 app
 ```
-or to enable both the MediaTek and QTI backends:
+or to enable multiple backends:
 ```
-make WITH_QTI=1 WITH_MEDIATEK=1 app
+make WITH_QTI=1 WITH_SAMSUNG=1 WITH_PIXEL=1 app
```
mcharleb, line 91: No-one but MLCommons can build the Samsung backend. Is it clear you would need to get the Samsung SDK (and Qualcomm SDK) to run this command? Why not WITH_MEDIATEK=1? Neuron delegate is not (and should not be) built by default.

mcharleb (changes requested):

anhappdev:
> @anhappdev Is there an issue associated with this PR? We typically try to create issues for each PR, and then review the issue list vs reviewing the PRs.

Yes. It's the issue #125 in the first comment.

android/README.md
anhappdev: I just wanted to show an example command to build multiple backends as stated in the text above, we can also add another text stated that building other backends require further setup described by each vendor.
Or we can leave the example command unchanged.

anhappdev: BTW, if Samsung backend is not public, should we remove the folder `mobile_back_samsung` to avoid confusion for public users?

mcharleb: The glue code is public but SDK and Makefile are only available under NDA

mcharleb: Only glue code (excluding Makefile) is public

anhappdev:
> We should have a separate issue that is closed by merging this PR (subset of #125)

Ok. I created the issue #135.

mcharleb (approved):
Looks good to me. I think @relja128 wanted to split the legacy app build into a android/app.mk and flutter app into flutter/app.mk which would then be included in the to level Makefile. That is what my other PR is doing but I can rebase on these changes.

PR #141, opened 2021-11-18
Add TOC and related repositories to README

### LINKED ISSUE #99: Central place for documentation
I think we need to have a central place for documentation or some solution to maintain documentation better.
The instructions are for example kind of copy & paste all over different repos and make it difficult to update.
There are many broken links and outdated instructions.
If someone has experience regarding this, please suggest a solution.

anhappdev:
Closes #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,12 @@
 This project contains the MLPerf mobile app, an app-based implementation of
 [MLPerf Inference](https://github.com/mlperf/inference) tasks.
 
+**TOC**:
+
+* [Overview](#overview)
+* [Folder Structure](#folder-structure)
+* [Related Repositories](#related-repositories)
+
 ## Overview
 
 The MLPerf app offers a simple mobile UI for executing MLPerf inference tasks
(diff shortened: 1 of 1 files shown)


mcharleb (approved):

mcharleb:
@anhappdev @d-uzlov  Do we know why CI is failing? Its blocking merges

d-uzlov:
@mcharleb, UI Crawl Test check is unstable because of some internal issues in firebase, Pixel 4 device often refuses to work.
I'll update the check to use different device.

README.md
```diff
@@ -3,6 +3,12 @@
 This project contains the MLPerf mobile app, an app-based implementation of
 [MLPerf Inference](https://github.com/mlperf/inference) tasks.
 
+**TOC**:
```
d-uzlov, line 6: Maybe it would be better to spell it fully, without abbreviations?
And also have it be a proper header, with `##`, instead of just bold text?

anhappdev: Yes. We can use the full text `Table of Contents`.
But bold text is better since TOC is not a section and does not have the same level of meaning like the other sections.

d-uzlov (approved):

Issue #142, opened 2021-11-19
Flutter app doesn't check error message from backend match


d-uzlov:
We don't check anywhere:
https://github.com/mlcommons/mobile_app_open/blob/817505143a9374ac99096c16057da1eadf21b7d3/flutter/cpp/flutter/dart_backend_match.cc#L27
https://github.com/mlcommons/mobile_app_open/blob/817505143a9374ac99096c16057da1eadf21b7d3/flutter/lib/backend/native.dart#L175

PR #143, opened 2021-11-19
Fix flutter backend match check

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/142

### DIFF
--- a/flutter/cpp/flutter/dart_backend_match.cc
+++ b/flutter/cpp/flutter/dart_backend_match.cc
@@ -8,10 +8,11 @@
 
 extern "C" struct dart_ffi_backend_match_result *dart_ffi_backend_match(
     const char *lib_path, const char *manufacturer, const char *model) {
-  LOG(INFO) << "checking backend '" << lib_path << "'...";
+  if (strlen(lib_path) == 0) LOG(INFO) << "checking built-in backend...";
+  else LOG(INFO) << "checking backend '" << lib_path << "' ...";
   ::mlperf::mobile::BackendFunctions backend(lib_path);
   if (*lib_path != '\0' && !backend.isLoaded()) {
-    LOG(INFO) << "backend '" << lib_path << "' can't be loaded";
+    LOG(INFO) << "backend can't be loaded";
     return nullptr;
   }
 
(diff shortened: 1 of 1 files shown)


flutter/cpp/flutter/dart_backend_match.cc
```diff
@@ -8,9 +8,10 @@
 
 extern "C" struct dart_ffi_backend_match_result *dart_ffi_backend_match(
     const char *lib_path, const char *manufacturer, const char *model) {
+  LOG(INFO) << "checking backend '" << lib_path << "'...\n";
   ::mlperf::mobile::BackendFunctions backend(lib_path);
   if (*lib_path != '\0' && !backend.isLoaded()) {
-    std::cerr << "backend '" << lib_path << "' can't be loaded\n";
+    LOG(INFO) << "backend '" << lib_path << "' can't be loaded\n";
```
anhappdev, line 14: The new line character here (and in other lines where it is appended to `LOG`) creates an empty line.

flutter/cpp/flutter/dart_backend_match.cc
```diff
@@ -8,9 +8,10 @@
 
 extern "C" struct dart_ffi_backend_match_result *dart_ffi_backend_match(
     const char *lib_path, const char *manufacturer, const char *model) {
+  LOG(INFO) << "checking backend '" << lib_path << "'...\n";
```
anhappdev, line 11: On iOS, there is only TFLite as default backend and the log shows
`I flutter/cpp/flutter/dart_backend_match.cc:11] checking backend ''...`
`I flutter/cpp/flutter/dart_backend_match.cc:67] backend '' matches`
Should we check for this case and make the log more understandable?

flutter/cpp/flutter/dart_backend_match.cc
d-uzlov: Fixed

flutter/cpp/flutter/dart_backend_match.cc
d-uzlov: Fixed

PR #151, opened 2021-11-25
Add Docker image to run code formatter

### LINKED ISSUE #148: Create Docker image for code formatting
We should have a Docker image contains all the tools needed to run code formatter, and a short guide how to use it.
All the code formatting tools and chosen code styles should match those used in our CI/CD pipeline.

anhappdev:
Closes #148


tools/formatter/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+# Set timezone to UTC by default
+RUN ln -sf /usr/share/zoneinfo/Etc/UTC /etc/localtime
+
+ENV FORMATTER_HOME /opt/formatters/bin
+ENV PATH=${PATH}:${FORMATTER_HOME}
+RUN mkdir -p ${FORMATTER_HOME}
+
+# Add Java formatter
+RUN apt-get install -y openjdk-11-jdk-headless
```
d-uzlov, line 34: It is generally recommended to run `apt-get update && apt-get install` everywhere, instead of just `install`
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

tools/formatter/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+FROM ubuntu:20.04
+
+ENV DEBIAN_FRONTEND="noninteractive"
+
+# Install dev packages
+RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-recommends \
+    tzdata gcc make git curl wget zip unzip tar gzip
+
+# Set timezone to UTC by default
+RUN ln -sf /usr/share/zoneinfo/Etc/UTC /etc/localtime
```
d-uzlov, line 27: Why do we need this?

tools/formatter/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+ENV FORMATTER_HOME /opt/formatters/bin
+ENV PATH=${PATH}:${FORMATTER_HOME}
+RUN mkdir -p ${FORMATTER_HOME}
+
+# Add Java formatter
+RUN apt-get install -y openjdk-11-jdk-headless
+RUN curl --output `dirname ${FORMATTER_HOME}`/google-java-format-1.9-all-deps.jar \
+         -L https://github.com/google/google-java-format/releases/download/google-java-format-1.9/google-java-format-1.9-all-deps.jar
+ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
```
d-uzlov, line 37: Is this really needed?
I believe that `apt-get install -y openjdk-11-jdk-headless` command should set this automatically.

tools/formatter/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+# Add buildifier
+RUN apt-get install -y golang-1.16-go
+ENV PATH=$PATH:/usr/lib/go-1.16/bin:~/go/bin
+RUN /usr/lib/go-1.16/bin/go get github.com/bazelbuild/buildtools/buildifier
+RUN cp ~/go/bin/buildifier ${FORMATTER_HOME}/
```
d-uzlov, line 43: What is the reason to copy binaries to FORMATTER_HOME? They are already in the PATH, right?

tools/formatter/format.mk
```diff
@@ -74,3 +74,19 @@ lint/line-endings:
 		echo -e "found files with CRLF line endings: \n$$_files"; false; \
 	else echo all files have unix line endings; \
 	fi
+
+.PHONY: docker/build/format
+docker/build/format:
```
d-uzlov, line 79: Can we make this a dependency for the format command, similar to how we treat Android app docker file, instead of making a user build it manually?
https://github.com/mlcommons/mobile_app_open/blob/c7c35f2ea73f3fad22755df7d3895802757c2c58/Makefile#L42
https://github.com/mlcommons/mobile_app_open/blob/c7c35f2ea73f3fad22755df7d3895802757c2c58/Makefile#L75

tools/formatter/Dockerfile
anhappdev: Fixed

tools/formatter/Dockerfile
anhappdev: Not in this case.

tools/formatter/Dockerfile
anhappdev: Fixed

tools/formatter/Dockerfile
anhappdev: I have no idea why but I need to copy it there else I have error command not found.

tools/formatter/format.mk
anhappdev: Fixed

tools/formatter/format.mk
```diff
@@ -74,3 +74,21 @@ lint/line-endings:
 		echo -e "found files with CRLF line endings: \n$$_files"; false; \
 	else echo all files have unix line endings; \
 	fi
+
+output/docker_mlperf_formatter.stamp: tools/formatter/Dockerfile
+	docker build --progress=plain \
+		--build-arg UID=`id -u` --build-arg GID=`id -g` \
+		-t mlperf/formatter tools/formatter
+	# need to clean flutter cache first else we will have error when running `dart run import_sorter:main` later in docker
+	cd flutter && flutter clean
```
d-uzlov, line 83: If we need this, we probably need this before each `docker/format` command.
We only build image 1 time, and something can invalidate Flutter cache later, when we already have formatter docker .stamp file.

anhappdev: We should not clear cache on every run. In my test, running `flutter clean` only one time there is enough. Subsequent run `make docker/format` or `flutter run` worked fine.

d-uzlov: Theoretically we shouldn't have to clear Flutter cache. In practice sometimes we do, but it doesn't have anything to do with building docker image.

I personally experience issues that are resolved by `flutter clean` when I switch between platforms. For example, if I use Flutter in Windows, and then launch Ubuntu in WSL and use mounted flutter folder there. Or something like this, I didn't try to find the exact conditions and commands.

If you got errors when running flutter in docker first time, then you can potentially get the same error next time when you switch between local flutter and docker.

anhappdev: I didn't try to find out why `flutter clean` is needed the first time when using `make docker/format`. But as stated, I can run `make docker/format` to run flutter in Docker container, or run `flutter run` on my Mac, and switch between 2 without any problem.

d-uzlov: Ok, mentioning the clean command in the documentation should be enough.

d-uzlov (approved):

tools/formatter/format.mk
anhappdev: If the problem may occur, then it is not worth to clear cache by every run. I added a note about clearing flutter cache in  `CONTRIBUTING.md`.

Issue #155, opened 2021-12-01
Flutter app doesn't pass native lib location to backend


d-uzlov:
See here for details: https://github.com/mlcommons/mobile_app_open/issues/111#issuecomment-982506436

In native Android app we can get this path using `getContext().getApplicationInfo().nativeLibraryDir`.
In the Flutter app we can't access `nativeLibraryDir` from the Dart code. I wasn't able to find any Flutter plugins that would allow this.
We can obtain this value in the Java/Kotlin code and transfer it to Dart using Flutter Method Channels.

PR #156, opened 2021-12-02
Fix empty native lib path in Flutter app

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/155

### DIFF
--- a/flutter/lib/backend/native.dart
+++ b/flutter/lib/backend/native.dart
@@ -1,9 +1,10 @@
 import 'dart:ffi';
 import 'dart:io';
 
+import 'package:flutter/services.dart';
+
 import 'package:device_info/device_info.dart';
 import 'package:ffi/ffi.dart';
-import 'package:flutter/services.dart';
 
 import 'package:mlcommons_ios_app/protos/backend_setting.pb.dart' as pb;
 import 'package:mlcommons_ios_app/protos/mlperf_task.pb.dart' as pb;
--- a/flutter/lib/backend/native.dart
+++ b/flutter/lib/backend/native.dart
@@ -3,10 +3,16 @@ import 'dart:io';
 
 import 'package:device_info/device_info.dart';
 import 'package:ffi/ffi.dart';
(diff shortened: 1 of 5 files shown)


anhappdev (commented):
The code looks good to me. Though, I have no suitable device to validate if the vendor backend works correctly.

flutter/lib/backend/native.dart
```diff
@@ -1,13 +1,20 @@
...
 import 'package:device_info/device_info.dart';
 import 'package:ffi/ffi.dart';
 
 import 'package:mlcommons_ios_app/protos/backend_setting.pb.dart' as pb;
 import 'package:mlcommons_ios_app/protos/mlperf_task.pb.dart' as pb;
 import 'bridge.dart';
 
+const androidChannel = MethodChannel('org.mlcommons.mlperfbench/android');
+Future<String> getNativeLibraryPath() async {
+  return await androidChannel.invokeMethod('getNativeLibraryPath') as String;
```
mohitmundhragithub, line 15: will this work for other OS as well, like windows / iOS?

d-uzlov: This method will throw an exception if we try to run it on Windows/iOS.
We only call it if `Platform.isAndroid` returned true:
https://github.com/mlcommons/mobile_app_open/blob/dea36c31605311a4b98d9d1f43146e479ace4dbd/flutter/lib/benchmark/benchmark.dart#L630-L631

d-uzlov: I don't know of any equivalent of `nativeLibraryDir` for other platforms.
On iOS we don't use any custom dynamic libraries yet. And we only use tflite backend, and there are no plans on adding more backends, as far as I'm aware.
On Windows the package is simpler that on Android/iOS, and we just keep all `.dll` files near executable.

d-uzlov:
@mohitmundhragithub, do you have any more comments, question or requests regarding this change?
If not, could you approve this PR?

mohitmundhragithub (approved):

PR #157, opened 2021-12-05
Add clear cache button to settings screen

### LINKED ISSUE #154: User can clear cache from the setting screen
Part of #125 
Add a button to the setting screen to let the user clear the current cache. All downloaded files (include the default `tasks.pbtxt` file) will be deleted and downloaded again.

anhappdev:
Closes #154.
Part of #125.


flutter/lib/ui/settings_screen.dart
```diff
@@ -146,6 +146,17 @@ class _SettingsScreen extends State<SettingsScreen> {
               }
             },
           ),
+          ListTile(
+            title: Text(
+              stringResources.clearCache,
+              style: TextStyle(color: Colors.blue),
```
d-uzlov, line 152: I think it's confusing that element that looks like just blue text and is placed near other text elements in settings acts like a button.
Can we use a real button?

anhappdev: I oriented to iOS Settings app, which uses blue text for reset action.
But I changed to use a `TextButton` now. I also added a confirm dialog.

d-uzlov (approved):

Issue #112, opened 2021-11-02
Merge Android and Flutter makefiles


d-uzlov:
We need to have uniform integrated makefiles across the repo.
Makefiles from the Flutter app repo don't match the style used in this repo, so they need to be significantly rewritten.

Parent issue: https://github.com/mlcommons/mobile_app_open/issues/92

d-uzlov:
@mcharleb, is there a particular reason why we use `@` everywhere in all makefiles to suppress the command echoing?

relja128:
Top level make files should be minimal, include app-specific make files from the subdirs.

anhappdev:
If the Android app will be deprecated in favor of Flutter app. I think it's best not to merge them. Just move the Makefile for Android into `android` folder.

d-uzlov:
I think that we should do the following in context of this task:

1. Pull make targets for the Flutter app to the repository root (and refactor them in the process)
2. Integrate all backends with the Flutter app
3. Add commands to build Flutter app with docker
4. And because all this would be too much for one makefile: refactor our root make file, moving Android-app-related stuff to the `/android` directory.

I also wanted to change the way we handle docker. Now there are just monstrous commands for everything in docker. I wanted to split each command (ones that we have now and ones that we will add in the future) into 2 parts: the `make command` itself, which builds what is needed locally, and `make docker/command`, that mounts docker container and runs `make command` inside.

And we should probably create a bunch more issues which would be linked to this one, and this one should be used just for tracking the status of the refactoring.

d-uzlov:
>If the Android app will be deprecated in favor of Flutter app

Another thing to note is that while the Android app itself will be deprecated, at least half of the configs and build commands (which were previously used for the Android app) are actually for backends, which are not going to be deprecated any time soon.

anhappdev:
>Another thing to note is that while the Android app itself will be deprecated, at least half of the configs and build commands (which were previously used for the Android app) are actually for backends, which are not going to be deprecated any time soon.

I would just copy those configs and command to some place outside the `android` folder. We should not have dependency on deprecated code. Sometime in the feature we will probably mark the `android` folder as deprecated and do not want to touch it anymore.

anhappdev:
@d-uzlov I think we can do the first point of #153 in this issue.

d-uzlov:
@anhappdev, which exact common resources are you talking about?

anhappdev:
> which exact common resources are you talking about?

[...]

PR #158, opened 2021-12-07
Add CI job for Flutter Android App

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/110


d-uzlov:
@anhappdev I disabled docker buildkit for Flutter CI because for some reason it often partially rebuilds the image even when nothing changed in the dockerfile and in docker build context.
You can go to the list of avaiable tags for `mlperf-mobile-android` image in Google Console and see that most of the images don't have any tags. This happened because dockerfile with the same git hash was rebuilt several times, and the tag containing this has moved to another image.
Do you know anything about this and how to fix it?

anhappdev:
Which docker version is used by the CI? 
It seems to be a bug with docker:
https://github.com/moby/buildkit/issues/1981
There is a comment that he still have the issue with the newer version, though it should be fixed.

d-uzlov:
Version of docker in CGP is 19.03.9, so we probably see this bug you linked.

flutter/Makefile
```diff
@@ -48,6 +46,33 @@ debug_flags_windows=-c dbg --copt /Od --copt /Z7 --linkopt -debug
...
+	@# take results from build/app/outputs/apk/androidTest/debug/app-debug-androidTest.apk
+	cd android && ./gradlew app:assembleAndroidTest
+	@# take results from build/app/outputs/apk/debug/app-debug.apk
+	cd android && ./gradlew app:assembleDebug -Ptarget=integration_test/first_test.dart
+
+.PHONY: docker/android/apk
+docker/android/apk: android/docker/image
+	@# if the build fails with java.io.IOException: Input/output error
+	@# remove file android/gradle/wrapper/gradle-wrapper.jar
+	MSYS2_ARG_CONV_EXCL="*" docker run --rm -it --user `id -u`:`id -g` -v $(CURDIR)/..:/mnt/project mlcommons/mlperf_mobile_flutter bash -c "cd /mnt/project/flutter && make prepare-flutter && make android/apk"
```
d-uzlov, line 66: This command doesn't utilize bazel cache.
I consider this command a proof that the docker container works both on CI and locally, and not a command that people should actually use to build the app.
A plan to add cache support while refactoring makefiles for https://github.com/mlcommons/mobile_app_open/issues/112.

flutter/Makefile
```diff
@@ -135,15 +160,15 @@ windows/flutter-release:
 
 .PHONY: backend-bridge-windows
 backend-bridge-windows:
-	bazel build ${_bazel_links_arg} --config=windows -c opt //flutter/cpp/flutter:backend_bridge.dll
+	bazel build ${BAZEL_CACHE_ARG} ${_bazel_links_arg} --config=windows -c opt //flutter/cpp/flutter:backend_bridge.dll
 	cd .. && chmod +w ${BAZEL_LINKS_DIR}bin/flutter/cpp/flutter/backend_bridge.dll
```
anhappdev, line 164: `BAZEL_CACHE_ARG` is passed from environment and `BAZEL_LINKS_DIR` is defined in the `Makefile`. I would somehow have a naming convention for the different. Since all other vars defined in the `Makefile` are lower-case. Maybe we can make `BAZEL_LINKS_DIR` lower-case?

flutter/Makefile
```diff
@@ -29,9 +29,7 @@ _bazel_links_arg=--symlink_prefix ${BAZEL_LINKS_DIR} --experimental_no_product_n
 .PHONY: cpp-ios
 cpp-ios:
 	@# NOTE: add `--copt -g` for debug info (but the resulting library would be 0.5 GiB)
-	bazel \
-		${BAZEL_CACHE_FLAG} \
-		build --config=ios_fat64 -c opt //flutter/cpp/flutter:ios_backend_fw_static
+	bazel ${BAZEL_OUTPUT_ROOT_ARG} build --config=ios_fat64 -c opt //flutter/cpp/flutter:ios_backend_fw_static
```
anhappdev, line 32: Where will `BAZEL_OUTPUT_ROOT_ARG` be set?

anhappdev (changes requested):

flutter/Makefile
d-uzlov: It is set in the iOS Jenkins job.
This is the main and only reason we had `BAZEL_CACHE_FLAG`.
The name was confusing to me so I renamed the env.

flutter/Makefile
d-uzlov: I used upper case for variables that I expect to be changed via environment, and lower case for variables that should never be changed outside of the makefile.
Default value for `BAZEL_LINKS_DIR` is defined in the makefile, but it is expected that user can change it.
Particularly, I added this because I needed to change it when building in Windows container, in `windows/docker/native` make target.

anhappdev (approved):

Issue #163, opened 2021-12-08
Move make targets to separate files included into root makefile


d-uzlov:
Part of https://github.com/mlcommons/mobile_app_open/issues/112
Depends on:
* https://github.com/mlcommons/mobile_app_open/issues/165

What we need to do:

1. Move contents of the root makefile to `/android` folder, and include it in root
2. Include makefile from `/flutter` into root makefile
3. Update documentation with new make targets

PR #167, opened 2021-12-09
Move make targets to separate files included into root makefile

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/163


d-uzlov:
It seems there are issues with macincloud service. Our macOS node is down, so CI for iOS doesn't work.

anhappdev (approved):

PR #170, opened 2021-12-11
Add offline mode to check for internet or local resources

### LINKED ISSUE #168: Add toggle switch to choose between internet resources or local resources.
Ref: https://github.com/mlcommons/mobile_app_open/issues/125#issuecomment-987601695

Add toggle switch to choose between internet resources and local resources.
There will be two configuration files, one with only local paths and one with only HTTP paths.
Local resources and HTTP resources will be stored separated in different folders.

anhappdev:
Closes #168 
Part of #125


d-uzlov:
@anhappdev I think I may have messed up the CI when I edited it for my last PR.
I adjusted iOS CI config a bit. Could you please push one more commit to see if I identified the issue correctly?

d-uzlov:
I'm sorry, my fix for CI config had a small error.
Could you push another commit?

d-uzlov (approved):

anhappdev:
@Mostelk Please review the PR if it matched your requirement.

d-uzlov:
Here is APK built with this PR: https://drive.google.com/file/d/1qjOVmP6wugZ9f1NIMHkdtPQ0u1LqBUuw/view?usp=sharing
Local custom config without any internet resources is required when using offline mode. Description how to create it is here: https://github.com/mlcommons/mobile_app_open/blob/e87636146ca82c6963296926d8cdc462550b5e16/flutter/documentation/custom-tasks.md#using-custom-taskspbtxt-file

Actually, I have just realized that this PR requires using a special custom backend, because our backends also require internet resources.
The APK I linked above includes the default backend, so it will never work in offline mode.
I guess this is not what we actually want. Am I right?

@Mostelk, does this look right?

flutter/lib/benchmark/resource_manager.dart
```diff
@@ -374,20 +374,24 @@ class ResourceManager {
...
-        benchmarksConfigurations.add(benchmarksConfiguration);
-      }
+    final benchmarksConfigurations = <BenchmarksConfiguration>[];
+
+    if (!await file.exists()) {
+      print('Create ' + file.path);
+      var config = <String, String>{
+        defaultBenchmarksConfiguration.name: defaultBenchmarksConfiguration.path
+      };
+      await file.writeAsString(jsonEncode(config));
```
d-uzlov, line 384: Function `getAvailableBenchmarksConfigurations` is only called when someone opens the list of available configurations.
Could you move this code to initialization stage?
I think `BenchmarkState::create` function would be convenient for this.

flutter/lib/benchmark/benchmark.dart
```diff
@@ -233,6 +233,33 @@ class BenchmarkState extends ChangeNotifier {
         datasetsError.map((element) => '\n${++index}) $element').join();
   }
 
+  Future<String> validateOfflineMode(String errorDescription) async {
+    if (!_store.offlineMode) {
+      return '';
+    }
+    final errors = <String>[];
+    for (final job in _getBenchmarkJobs()) {
+      final modelPath = job.benchmark.benchmarkSetting.src;
```
d-uzlov, line 242: Do we really want to check model path?
It is defined by the backend, and there is nothing we can do to change it except to edit sources and rebuild the app.
I though that this offline mode check was about making sure that we use the appropriate `tasks` file, while using the release binary.

flutter/lib/benchmark/benchmark.dart
```diff
@@ -233,6 +233,33 @@ class BenchmarkState extends ChangeNotifier {
...
+      if (isInternetResource(modelPath)) {
+        errors.add(modelPath);
+      }
+      final testDataPath = job.dataset.groundtruthSrc;
+      if (isInternetResource(testDataPath)) {
+        errors.add(testDataPath);
+      }
+      final groundtruthDataPath = job.dataset.path;
```
d-uzlov, line 250: Typo: variables names are switched in lines 246 and 250.

d-uzlov (changes requested):

flutter/lib/benchmark/benchmark.dart
anhappdev: From my understanding that some other engineers may build the app on his own and use his models to test.

@Mostelk Can you clarify?

flutter/documentation/custom-tasks.md
```diff
@@ -9,11 +9,19 @@ This file describes how to configure the application to use non-default benchmar
...
+When the app starts for the first time, it creates a file named `benchmarksConfigurations.json` 
+in the root application directory.
+Each entry in the `benchmarksConfigurations.json` is the name and location of a configuration file.
+You can have multiple configurations stored in the `benchmarksConfigurations.json` and switch 
+between them in the app settings. 
+The `benchmarksConfigurations.json` looks like this:
+```json
+{
+   "default":"https://raw.githubusercontent.com/mlcommons/mobile_models/main/v1_0/assets/tasks_v2.pbtxt",
+   "otherName":"/sdcard/location/to/tasks.pbtxt"
```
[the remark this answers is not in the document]
anhappdev: Fixed

flutter/lib/benchmark/resource_manager.dart
anhappdev: Good catch. Thanks.

flutter/lib/benchmark/benchmark.dart
anhappdev: Fixed

flutter/documentation/custom-tasks.md
```diff
@@ -9,13 +9,23 @@ This file describes how to configure the application to use non-default benchmar
...
+When the app starts for the first time, it creates a file named `benchmarksConfigurations.json` 
+in the root application directory.
+Each entry in the `benchmarksConfigurations.json` is the name and location of a configuration file.
+You can have multiple configurations stored in the `benchmarksConfigurations.json` and switch 
+between them in the app settings. 
+The `benchmarksConfigurations.json` looks like this:
+```json
+{
+   "default":"https://raw.githubusercontent.com/mlcommons/mobile_models/main/v1_0/assets/tasks_v2.pbtxt",
+   "otherName":"app:///path/to/custom_tasks.pbtxt"
```
d-uzlov, line 21: Only filename matters when specifying config file path.
See here: https://github.com/mlcommons/mobile_app_open/blob/master/flutter/lib/benchmark/benchmark.dart#L351-L353
App would throw out `app:///path/to/` or any other prefix and look specifically for `<app dir>/custom_tasks.pbtxt`

flutter/lib/benchmark/resource_manager.dart
```diff
@@ -136,6 +136,17 @@ class ResourceManager {
     await Directory(loadedResourcesDir).create();
   }
 
+  Future<void> createConfigurationFile() async {
+    final file = File('$applicationDirectory/$_configurationsFileName');
+    if (!await file.exists()) {
+      print('Create ' + file.path);
+      var config = <String, String>{
+        defaultBenchmarksConfiguration.name: defaultBenchmarksConfiguration.path
```
d-uzlov, line 144: If we update the app and default configuration URL changes, but we already have the `benchmarksConfigurations.json` file, the app will not use the new URL unless we clear app data or manually delete `benchmarksConfigurations.json`.
Is this the expected behavior?

flutter/documentation/custom-tasks.md
anhappdev: There is a check for HTTP prefix some line above, so the app would just use the config file path as is in this case. I fixed it by parsing the path first some line under if it is not an HTTP link.

flutter/lib/benchmark/resource_manager.dart
anhappdev: I remember there was a discussion how we should handle updated default configuration. There are cases:
1. The URL is changed but app is not changed.
2. The URL is not changed but the content of the config file is changed.

This would be best to handle in a separated issue.

d-uzlov: Could you open a new issue for this after we merge this PR?

anhappdev: https://github.com/mlcommons/mobile_app_open/issues/184

d-uzlov (approved):
@anhappdev, we likely won't get any feedback on this for a few weeks.
I think it would be better to merge it now to avoid merge conflicts, and then fix any further issues separately.

PR #176, opened 2021-12-18
Change parsing logic for local resources in tasks.pbtxt

### LINKED ISSUE #171: Change parsing logic of location path in tasks.pbtxt
Currently, the Flutter app will replace `mlperf_datasets` with `external_resources`. See [this README](https://github.com/mlcommons/mobile_app_open/blob/0e161989a7d17456f78c5c83708b9250ce81a83d/flutter/documentation/custom-tasks.md?plain=1#L27).
This seems to be a workaround to make the Flutter app compatible with the [tasks_v2.pbtxt](https://github.com/mlcommons/mobile_models/blob/main/v1_0/assets/tasks_v2.pbtxt) used in the Android app.

This is quite confusing because if the path defined the `tasks.pbtxt` is `/sdcard/mlperf_datasets/some/folders/test.txt` then the path on iPhone is `/MLPerf/external_resources/some/folders/test.txt`.

We should remove this logic and use the location defined in `tasks.pbtxt` as is.

anhappdev:
Close #171.

Old logic was replacing `/sdcard/mlperf_datasets/file.txt` with `<app_dir>/external_resources/file.txt`.
New logic is now replacing `app:///mlperf_datasets/file.txt` with `<app_dir>/mlperf_datasets/file.txt`.
That means the path for local resources in `tasks.pbtxt` must be prefixed with `app://`

### DIFF
--- a/flutter/lib/benchmark/resource_manager.dart
+++ b/flutter/lib/benchmark/resource_manager.dart
@@ -30,15 +30,18 @@ class BatchPreset {
 
 class ResourceManager {
   static const _zipPattern = '.zip';
-  static const _systemPathPattern = '/sdcard/mlperf_datasets';
+  static const _applicationDirectoryPrefix = 'app://';
   static const _configurationsFileName = 'benchmarksConfigurations.json';
+  static const _jsonResultFileName = 'result.json';
+  static const _loadedResourcesDirName = 'loaded_resources';
 
   final VoidCallback onUpdate;
   final HttpClient _httpClient = HttpClient();
   final BenchmarksConfiguration defaultBenchmarksConfiguration =
       BenchmarksConfiguration(
     'default',
-    'https://raw.githubusercontent.com/mlcommons/mobile_models/main/v1_0/assets/tasks_v2.pbtxt',
+    // [anh] TODO: Replace this temp URL before merge PR
+    'https://www.dropbox.com/s/kx3qbe524oj95g1/171_tasks_v3.pbtxt?dl=1',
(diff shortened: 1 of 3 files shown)


flutter/lib/benchmark/resource_manager.dart
```diff
@@ -30,15 +30,18 @@ class BatchPreset {
...
+  static const _jsonResultFileName = 'result.json';
+  static const _loadedResourcesDirName = 'loaded_resources';
 
   final VoidCallback onUpdate;
   final HttpClient _httpClient = HttpClient();
   final BenchmarksConfiguration defaultBenchmarksConfiguration =
       BenchmarksConfiguration(
     'default',
-    'https://raw.githubusercontent.com/mlcommons/mobile_models/main/v1_0/assets/tasks_v2.pbtxt',
+    // [anh] TODO: Replace this temp URL before merge PR
```
mohitmundhragithub, line 43: This still needs to be replaced?

mohitmundhragithub (approved):

flutter/lib/benchmark/resource_manager.dart
anhappdev: Yes. Should have marked this PR as draft.
We need to agreed on this new prefix `app:\\` for local path. Then I will create a new `tasks.pbtxt` file, push it to repo `mobile_models`, replace this link before merge.

d-uzlov (approved):
The code and documentation changes LGTM.

PR #181, opened 2021-12-22
fix neuron delegate batch size logic

### LINKED ISSUE #180: Flutter app doesn't really use Neuron delegate
From `logcat`, Neuron backend is detected but NNAPI delegate is used.

freedomtan:
fix a batch size checking issue (it seems the issue was there when tasks schema changed couple months ago). Close #180

### DIFF
--- a/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
+++ b/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
@@ -260,7 +260,7 @@ mlperf_backend_ptr_t mlperf_backend_create(
     } else if (strcmp(configs->accelerator, "neuron") == 0) {
       // The kOptimizationBatchProcessor doesn't work yet.
       // Use NNAPI instead.
-      if (configs->batch_size != 1) {
+      if (configs->batch_size > 1) {
         auto options = tflite::StatefulNnApiDelegate::Options();
         options.allow_fp16 = true;
         options.disallow_nnapi_cpu = true;
--- a/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
+++ b/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
@@ -267,7 +267,7 @@ mlperf_backend_ptr_t mlperf_backend_create(
       // to handle it.
       //
       // When batch_size > 1 and divisible by shards_num, we have
-      //   backend_data->real_batch_size = 
+      //   backend_data->real_batch_size =
       //     configs->batch_size / backend_data->shards_num;
       // which is not what we want. E.g., for shards_num = 2 and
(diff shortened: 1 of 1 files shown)


mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
```diff
@@ -260,7 +260,7 @@ mlperf_backend_ptr_t mlperf_backend_create(
     } else if (strcmp(configs->accelerator, "neuron") == 0) {
       // The kOptimizationBatchProcessor doesn't work yet.
       // Use NNAPI instead.
-      if (configs->batch_size != 1) {
+      if (configs->batch_size > 1) {
```
d-uzlov, line 263: Shouldn't we use `backend_data->real_batch_size` here?

freedomtan: Nope, `backend_data->real_batch_size` also works for current hardware, but it's not what we want.

If I I have 4 NPUs which perform best with batch size one (4 shards/threads, real_batch_size == 1), I still want to use NNAPI delegate rather than Neuron delegate. 

By the way, it seems the batch_size (128 for neuron backend) in backend setting file is overwritten by numbers from UI (4 threads, batch size: 2).

d-uzlov: >Nope, backend_data->real_batch_size also works for current hardware, but it's not what we want.

Could you add a comment about this in the code, to make sure no one tries to "fix" it?

>By the way, it seems the batch_size (128 for neuron backend) in backend setting file is overwritten by numbers from UI (4 threads, batch size: 2).

Yep. If I remember correctly, Flutter app didn't use batch size at all at first, and when we fixed it, we were optimizing it for iPhones.
We should probably rework this behavior at some point, maybe in January.

anhappdev (approved):
LGTM

mobile_back_tflite/cpp/backend_tflite/tflite_c.cc
```diff
@@ -260,7 +260,12 @@ mlperf_backend_ptr_t mlperf_backend_create(
     } else if (strcmp(configs->accelerator, "neuron") == 0) {
       // The kOptimizationBatchProcessor doesn't work yet.
       // Use NNAPI instead.
-      if (configs->batch_size != 1) {
+      //
+      // Here we use the batch size of the task to check if this
+      // is for offline/batch scenario, if batch size is not
+      // larger than 1, assume it's offline and use NNAPI delegate
+      // to handle it.
```
d-uzlov, line 267: It doesn't really help to understand **_why_** we use `configs->batch_size` instead of `backend_data->real_batch_size`.
Could you rephrase the comment to make it more clear?

d-uzlov (approved):

Issue #172, opened 2021-12-14
Refactor Flutter files


d-uzlov:
We have few files in the Dart/Flutter sources that are hard to read and navigate.
[benchmark.dart](https://github.com/mlcommons/mobile_app_open/blob/0e161989a7d17456f78c5c83708b9250ce81a83d/flutter/lib/benchmark/benchmark.dart) — 670 lines
[resource_manager.dart](https://github.com/mlcommons/mobile_app_open/blob/0e161989a7d17456f78c5c83708b9250ce81a83d/flutter/lib/benchmark/resource_manager.dart) — 477 lines
But the primary reason for the inconvenience is the lack of proper structure, in these and in some of the smaller files.
It would be nice to refactor it.

Estimated required time: ~~1-2 days~~ 1-2 weeks.

This is a low priority task.

d-uzlov:
Currently the only file that really needs refactoring is `flutter/lib/benchmark/state.dart`:
https://github.com/mlcommons/mobile_app_open/blob/40e01139b89e73acada30f0966f7d116ad7cdc7b/flutter/lib/benchmark/state.dart
Its current structure may significantly increase time needed to modify anything related to it. I haven't needed to adjust anything in it lately but we will definitely need to fix it at some point.

freedomtan:
Close it for now. May reopen it with more details later

PR #186, opened 2021-12-28
Refactor resource_manager.dart

d-uzlov:

This is a separate PR for a small part of the refactoring to make merge conflicts less likely and reviews easier.


flutter/lib/resources/archive_cache_helper.dart
```diff
@@ -0,0 +1,58 @@
+import 'dart:io';
+
+import 'package:archive/archive_io.dart';
+
+import 'file_cache_helper.dart';
+
+class ArchiveCacheHelper {
+  static const extension = '.zip';
+  final FileCacheHelper cacheManager;
```
anhappdev, line 9: The variable is named as `cacheManager` but has type `FileCacheHelper` and not `CacheManager`. Should be `fileCacheHelper`.

flutter/lib/resources/cache_manager.dart
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      resourcesToDownload.add(resource);
+
+      continue;
+    }
+    await _download(resourcesToDownload, reportProgress);
+
+    if (purgeOldCache) {
+      const atLeastDaysOld = 30 * 9;
```
anhappdev, line 93: Maybe we can make this `atLeastDaysOld` a global constant or at least class constant.

flutter/lib/resources/configurations_manager.dart
```diff
@@ -0,0 +1,75 @@
...
+    final jsonContent = await _readConfigs();
+    final configPath = jsonContent[name] as String?;
+
+    if (configPath != null) {
+      return BenchmarksConfig(name, configPath);
+    }
+    if (name == defaultConfig.name) {
+      return defaultConfig;
+    }
+    return null;
```
anhappdev, line 73: I think this part is redundant since we always create a config file if not existed and read from it.

flutter/lib/resources/resource_manager.dart
```diff
@@ -0,0 +1,166 @@
...
+    ));
+    result.add(BatchPreset(
+      name: 'Custom',
+      batchSize: 0,
+      shardsCount: 0,
+    ));
+
+    final yamlString =
+        await rootBundle.loadString('assets/default_batch_settings.yaml');
+    for (var item in loadYaml(yamlString)['devices']) {
```
anhappdev, line 145: I think it's better to make the string `assets/default_batch_settings.yaml` a static const.

flutter/lib/resources/result_manager.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:convert';
+import 'dart:io';
+
+class ResultManager {
+  static const _jsonResultFileName = 'result.json';
```
anhappdev, line 5: Do you have the file `result.json` in the app root folder on Android after running benchmark? I have one on iOS but not on Android.

anhappdev (changes requested):

flutter/lib/resources/result_manager.dart
```diff
@@ -0,0 +1,38 @@
...
+
+  Future<String> get json async {
+    final file = File(_jsonResultPath);
+
+    if (await file.exists()) {
+      return file.readAsString();
+    }
+
+    return '';
+  }
```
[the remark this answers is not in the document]
d-uzlov: It is meant to be used as `resultManager.json`:
https://github.com/mlcommons/mobile_app_open/blob/214521a9c5098820d7dec5a70bde7273a3beaeae/flutter/lib/ui/result_screen.dart#L361
I think the word `result` would be redundant here.
And I think it's hard to confuse `resultManager.json` as being json representation of the class.

I'm not a big fan of this "getter" syntax, though. Maybe it would be better to make it a function with name like `readJson`.
What do you think?

flutter/lib/resources/configurations_manager.dart
d-uzlov: This part is redundant, but for a different reason.
We don't control the `benchmarksConfigurations` file, it's possible that someone deleted the `default` entry from there. But we have another "use default if other configs are not found" logic in `benchmark.dart`.

flutter/lib/resources/result_manager.dart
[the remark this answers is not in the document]
anhappdev: Because of the `Manager` in the class name, it's not clear that `json` means the result.
I think `read()` would be enough. Like other methods `write()`, `delete()`.

flutter/lib/resources/result_manager.dart
d-uzlov: Yes, I see this file on my Android phone.

flutter/lib/resources/result_manager.dart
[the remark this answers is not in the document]
d-uzlov: fixed

flutter/lib/resources/archive_cache_helper.dart
d-uzlov: fixed

flutter/lib/resources/cache_manager.dart
d-uzlov: fixed

flutter/lib/resources/configurations_manager.dart
d-uzlov: fixed

flutter/lib/resources/resource_manager.dart
d-uzlov: fixed

anhappdev (approved):

PR #191, opened 2021-12-29
Refactor benchmark.dart

d-uzlov:

This is a separate PR for a small part of the refactoring to make merge conflicts less likely and reviews easier.

### DIFF
--- a/flutter/lib/resources/configurations_manager.dart
+++ b/flutter/lib/resources/configurations_manager.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:mlperfbench/localizations/app_localizations.dart';
+import 'package:mlperfbench/resources/resource_manager.dart';
 import 'utils.dart';
 
 const _configurationsFileName = 'benchmarksConfigurations.json';
@@ -21,10 +22,49 @@ class BenchmarksConfig {
 
 class ConfigurationsManager {
   final String applicationDirectory;
+  final ResourceManager resourceManager;
   final BenchmarksConfig defaultConfig =
       BenchmarksConfig('default', _defaultConfigUrl);
+  String currentConfigName;
+    String configFilePath = '';
 
-  ConfigurationsManager(this.applicationDirectory);
+  ConfigurationsManager(this.applicationDirectory, this.currentConfigName, this.resourceManager);
+
+  Future<BenchmarksConfig?> get currentConfig async =>
+      await getConfig(currentConfigName);
+
+      
+  Future<void> setConfig(BenchmarksConfig config) async {
+    if (isInternetResource(config.path)) {
+      configFilePath = await resourceManager.cacheManager.fileCacheHelper
+          .get(config.path, true);
+    } else {
+      configFilePath = resourceManager.get(config.path);
+      if (!await File(configFilePath).exists()) {
+        throw 'local config file is missing: $configFilePath';
+      }
+    }
+
+    if (currentConfigName == config.name) {
+      return;
+    }
+
+    currentConfigName = config.name;
+
+    final nonRemovableResources = <String>[];
+    if (isInternetResource(config.path)) {
+      nonRemovableResources.add(resourceManager.cacheManager.fileCacheHelper
+          .getResourceRelativePath(config.path));
+    }
+
+    await resourceManager.cacheManager
+        .deleteLoadedResources(nonRemovableResources);
+  }
+
+  String getConfigFilePath() {
+    return configFilePath;
+  }
+
 
   Future<File> createConfigurationsFile() async {
     final file = File('$applicationDirectory/$_configurationsFileName');
--- a/flutter/lib/resources/configurations_manager.dart
+++ b/flutter/lib/resources/configurations_manager.dart
@@ -26,7 +26,7 @@ class ConfigurationsManager {
   final BenchmarksConfig defaultConfig =
       BenchmarksConfig('default', _defaultConfigUrl);
   String currentConfigName;
-  String configFilePath = '';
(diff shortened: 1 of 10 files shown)


d-uzlov:
I think it's better to approach `benchmark.dart` in parts. It's the main state machine of the app and an example of "God object" antipattern, so a complete rework would be very hard to review.

In this PR I only pull out result and config (tasks file) manipulations out of the `BenchmarkState` class.

flutter/lib/resources/result_manager.dart
```diff
@@ -9,7 +50,71 @@ class ResultManager {
...
+        'min_duration': result.minDuration.toString(),
+        'duration': result.durationMs.toString(),
+        'min_samples': result.minSamples.toString(),
+        'num_samples': result.numSamples.toString(),
+        'shards_num': result.threadsNumber,
+        'batch_size': result.batchSize,
+        'mode': result.mode.toString(),
+        'datetime': DateTime.now().toIso8601String(),
+      };
+      var briefResult = _BriefResult(result.id, score, result.accuracy,
```
anhappdev, line 81: Since there is a `_BriefResult`, can you also create a `_BenchmarkResult`.

flutter/lib/ui/benchmarks_configuration_screen.dart
```diff
@@ -54,14 +50,14 @@ class BenchmarksConfigurationScreen extends StatelessWidget {
   Widget build(BuildContext context) {
     final stringResources = AppLocalizations.of(context);
     final state = context.watch<BenchmarkState>();
-    final chosenConfiguration = state.chosenBenchmarksConfiguration;
+    final currentConfiguration = state.configManager.currentConfig;
```
anhappdev, line 53: I would stick to the short form `Config` instead of `Configuration`

flutter/lib/resources/configurations_manager.dart
```diff
@@ -21,12 +22,46 @@ class BenchmarksConfig {
...
 class ConfigurationsManager {
   final String applicationDirectory;
+  final ResourceManager resourceManager;
   final BenchmarksConfig defaultConfig =
       BenchmarksConfig('default', _defaultConfigUrl);
+  String currentConfigName;
+  String configPath = '';
 
-  ConfigurationsManager(this.applicationDirectory);
+  ConfigurationsManager(
```
anhappdev, line 31: Maybe rename this class to `ConfigManager` to be consistent with using the short form.

anhappdev (changes requested):
Requested changes are mostly just naming. Else I think the refactoring is quite good. For example I did not like the method `Future<File?> handleChosenConfiguration({BenchmarksConfig? newChosenConfiguration, required Store store})` because it modifies state (side effect) and returns a value at the same time, but now it's gone.  👍

flutter/lib/resources/result_manager.dart
d-uzlov: What would be the benefit?
I created `_BriefResult` class because we parse it. It just doesn't look good when we define 2 equivalent string constants to write and then to parse json in different places in code, and use quite different logic for reading and writing, which can result in reading not what we thought we wrote. Looked kinda brittle.
We don't parse full results json anywhere, we just use for sharing. Creating a class for it would increase number of line but wouldn't really improve anything in my opinion.

Maybe we should just move json serialization of full results to a separate method?
Or we could just delete `_BriefResult` class and save results in the same format that we use for sharing.

anhappdev: Both `BriefResult` and `BenchmarkResult` are data models. I think it's best practice to define data model as a `class` (or `struct` depend on language). Methods for parsing/storing can be placed inside `ResultManager` or we can make both `BriefResult` and `BenchmarkResult` conform to a JSON serialization / deserialization protocol and `ResultManager` only handles the reading from disk or storing to disk.

flutter/lib/resources/configurations_manager.dart
```diff
@@ -38,11 +73,11 @@ class ConfigurationsManager {
   }
 
   Future<Map<String, dynamic>> _readConfigs() async {
-    final file = await createConfigurationFile();
+    final file = await createConfigurationsFile();
     return jsonDecode(await file.readAsString()) as Map<String, dynamic>;
   }
 
-  Future<List<BenchmarksConfig>> getAvailableConfigs() async {
+  Future<List<BenchmarksConfig>> getList() async {
```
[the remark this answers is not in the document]
d-uzlov: done

flutter/lib/resources/configurations_manager.dart
```diff
@@ -21,12 +22,46 @@ class BenchmarksConfig {
...
+    if (isInternetResource(config.path)) {
+      nonRemovableResources.add(resourceManager.cacheManager.fileCacheHelper
+          .getResourceRelativePath(config.path));
+    }
+
+    await resourceManager.cacheManager
+        .deleteLoadedResources(nonRemovableResources);
+  }
+
+  Future<File> createConfigurationsFile() async {
```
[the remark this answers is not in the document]
d-uzlov: done

flutter/lib/resources/result_manager.dart
d-uzlov: done

flutter/lib/ui/benchmarks_configuration_screen.dart
d-uzlov: done

flutter/lib/resources/configurations_manager.dart
d-uzlov: done

anhappdev (approved):

PR #194, opened 2021-12-29
Show backend used in benchmarks.

### LINKED ISSUE #178: Flutter UI should show the backend used in the advanced view just like the old app

anhappdev:
Closes #178


flutter/lib/ui/result_screen.dart
```diff
@@ -112,6 +112,8 @@ class _ResultScreenState extends State<ResultScreen>
     final pictureEdgeSize = 0.08 * MediaQuery.of(context).size.width;
 
     for (final benchmark in state.benchmarks) {
+      final backendDesc =
+          '${benchmark.benchmarkSetting.configuration} | ${benchmark.benchmarkSetting.acceleratorDesc}';
```
d-uzlov, line 116: Previous run could be made with different settings. For example, if someone reinstalls the app with different backend list, or with updated backend.
We should save the config info in the results and use it for showing previous results.

d-uzlov (changes requested):

d-uzlov:
I should note that tflite backend uses strategy "use CPU if specified delegate failed", and we use iOS settings for Windows, so now we can see `CoreML` labels in the benchmark list in Windows builds.
It's not the fault of this PR, but maybe we should do something about it in the future. Probably, create an issue to create a separate config for Windows.

flutter/lib/ui/result_screen.dart
anhappdev: Ok. I made the change to store the backend description in `_BriefResult` and `_FullResult`.
Because the `_BriefResult` is directly parsed from `RunResult` in `serializeBriefResults()`, I need to add `backendDescription` to both `RunResult` and `RunSettings`. I am not quite happy with this approach, since `backendDescription` is not used in the backend, but cannot find a cleaner approach. I am open to suggestion.

anhappdev: We now have a use case for `backendDescription` in the backend part, see https://github.com/mlcommons/mobile_app_open/issues/203#issue-1097907108

anhappdev:
> I should note that tflite backend uses strategy "use CPU if specified delegate failed", and we use iOS settings for Windows, so now we can see `CoreML` labels in the benchmark list in Windows builds.
> It's not the fault of this PR, but maybe we should do something about it in the future. Probably, create an issue to create a separate config for Windows.

I created a new issue for this https://github.com/mlcommons/mobile_app_open/issues/203#issue-1097907108

d-uzlov (approved):

Issue #30, opened 2021-08-11
Samsung backend does not destroy itself between tasks


bellycat77:
Original https://github.com/mlcommons/mobile_app/issues/610

relja128:
Per @Mostelk this is getting pushed back to post 1.1 (waiting for new submission)

relja128:
Per @mcharleb - This is the check in the front end, the FE doesn't delete the backend, because the BE used to crash on delete. Probably in taskrunner.
@Mostelk - will take a look

TheKanter:
@Mostelk When can we expect your team to clear this?

relja128:
Per @Mostelk - v2 should fix this.

al-mahi:
PR [197](https://github.com/mlcommons/mobile_app_open/pull/197)

al-mahi:
Updated PR https://github.com/mlcommons/mobile_app_open/pull/197

relja128:
Pushing this off for post 2.0 submission.

freedomtan:
@jwookiehong will check if the latest samsung backend resolved this issue.

AhmedTElthakeb:
Will need this [PR ](https://github.com/mlcommons/mobile_app_open/pull/271) to integrate our new and old backends

freedomtan:
It seems there is not such problem for the new Samsung backend. Close this for now.

AhmedTElthakeb:
A workaround is provided in the following [PR#283](https://github.com/mlcommons/mobile_app_open/pull/283), newer generations won't have this problem.

AhmedTElthakeb:
This issue has been resolved for all generations (old and new) in the latest android app merge, no workarounds are needed any more. 
https://github.com/mlcommons/mobile_app_open/tree/android-v2
Commit# a47423

PR #197, opened 2021-12-31
#30 samsungbackend close workaround

al-mahi:
driverWrapper.close() function call ends up calling external.cc files tflite::UnloadLibrary(handle) function which is essentially a dlclose() call on libsamsungbackend.so file. Moving the workaround of commenting driverWrapper.close() to deeper level is more specific to the issue. Also allocated memory dynamically for the backend object.

### DIFF
--- a/android/cpp/backends/external.cc
+++ b/android/cpp/backends/external.cc
@@ -79,7 +79,7 @@ BackendFunctions::BackendFunctions(const std::string& lib_path) {
     }
     handle = tflite::SharedLibrary::LoadLibrary(wide_lib_path.c_str());
 #else
-    handle = tflite::SharedLibrary::LoadLibrary(lib_path.c_str());
+    handle = dlopen(lib_path.c_str(), RTLD_LAZY | RTLD_LOCAL | RTLD_NODELETE);
 #endif
     if (handle == nullptr) {
       LOG(ERROR) << "Unable to load: " << lib_path << ": "
--- a/android/cpp/backends/external.h
+++ b/android/cpp/backends/external.h
@@ -31,7 +31,7 @@ struct BackendFunctions {
   BackendFunctions(const std::string& lib_path);
 
   ~BackendFunctions() {
-    if (handle) {
+    if (false) {
       tflite::SharedLibrary::UnLoadLibrary(handle);
     }
   }
--- a/android/cpp/backends/external.h
+++ b/android/cpp/backends/external.h
@@ -31,7 +31,7 @@ struct BackendFunctions {
   BackendFunctions(const std::string& lib_path);
 
   ~BackendFunctions() {
-    if (false) {
+    if (handle) {
       tflite::SharedLibrary::UnLoadLibrary(handle);
     }
   }
--- a/mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
+++ b/mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
@@ -1053,7 +1053,7 @@ void mlperf_backend_clear(mlperf_backend_ptr_t backend_ptr) {
   delete samsung_backend;
 }
 
-static Backend *ss_backend;
+static Backend* ss_backend;
 
 // Create a new backend and return the pointer to it.
 mlperf_backend_ptr_t mlperf_backend_create(
--- a/mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
+++ b/mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
@@ -1050,15 +1050,17 @@ void mlperf_backend_clear(mlperf_backend_ptr_t backend_ptr) {
   EdenAI::jni_CloseModel();
   EdenAI::jni_Shutdown();
   libeden_nn_on_direct = nullptr;
+  delete samsung_backend;
 }
 
-Backend ss_backend;
+static Backend *ss_backend;
 
 // Create a new backend and return the pointer to it.
 mlperf_backend_ptr_t mlperf_backend_create(
     const char* model_path, mlperf_backend_configuration_t* configs,
     const char* native_lib_path) {
-  Backend* samsung_backend = &ss_backend;
+  ss_backend = new Backend();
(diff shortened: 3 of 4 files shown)


anhappdev:
@al-mahi Please make sure all tests are passed. `Format & lint / C++` is failed.

android/cpp/backends/external.h
```diff
@@ -31,7 +31,7 @@ struct BackendFunctions {
   BackendFunctions(const std::string& lib_path);
 
   ~BackendFunctions() {
-    if (handle) {
+    if (false) {
```
anhappdev, line 34: This doesn't look right to me.

al-mahi: @anhappdev I am basically stopping tflite::SharedLibrary::UnLoadLibrary(handle); or dlclose(handle); call here as a workaround of the app crash.

anhappdev: 1. Yes, `if (false)` means the code in the if block will never be called, why not just remove it?
2. Does this change have an impact on other backends? I am not familiar with the backend code yet, but the code was there for a reason. So other engineers need to review this.

al-mahi:
@relja128  I have updated the PR please review again.
@mohitmundhragithub Would you please let us know if you find any issue while testing with Qualcomm backend.
@freedomtan  Would you please let us know if you find any issue while testing with MTK backend.
`// handle = tflite::SharedLibrary::LoadLibrary(lib_path.c_str()); <-- master branch`
`handle = dlopen(lib_path.c_str(), RTLD_LAZY | RTLD_NODELETE | RTLD_LOCAL);<-- potential fix or workaround`

Some of the changes were still useful from the PR like dynamic memory allocation so I have not created a new PR.  @anhappdev was concerned about the changes in exeternal.h file so I reverted it back and now I added a flag while opening the backend which prevent crash for samsung backend.

d-uzlov (changes requested):
Using `RTLD_NODELETE` is [equivalent](https://linux.die.net/man/3/dlopen) to deleting `dlclose`/`UnLoadLibrary`.
I guess the desired effect here is to skip reinitialization of static variables, or something?

Removing/disabling `dlclose` does not look good in my opinion.
Code quality aside, we will keep resources from all backends in memory, because all of them will be loaded to call `mlperf_backend_matches_hardware` but never unloaded. This can affect memory usage if some backends have large static variables, or this can potentially affect other things if some backends have non-trivial static variables.

We should either change the logic to "load the matched backend once and keep the handle forever" (which can be discussed separately) or fix the issue with static variables.

If you want to use a workaround of never reloading the backend, Samsung backend should be able to just `dlopen` itself from the inside (then after app does a `dlclose` there will still be open handles, so backend will not be unloaded). Our API doesn't pass library path to backend so you can't call `dlopen` directly, but you can use [dladdr](https://man7.org/linux/man-pages/man3/dladdr.3.html) to get path.

mohitmundhragithub:
> @relja128 I have updated the PR please review again. @mohitmundhragithub Would you please let us know if you find any issue while testing with Qualcomm backend. @freedomtan Would you please let us know if you find any issue while testing with MTK backend. `// handle = tflite::SharedLibrary::LoadLibrary(lib_path.c_str()); <-- master branch` `handle = dlopen(lib_path.c_str(), RTLD_LAZY | RTLD_NODELETE | RTLD_LOCAL);<-- potential fix or workaround`
> 
> Some of the changes were still useful from the PR like dynamic memory allocation so I have not created a new PR. @anhappdev was concerned about the changes in exeternal.h file so I reverted it back and now I added a flag while opening the backend which prevent crash for samsung backend.

it seems to be working fine for us.

mohitmundhragithub (approved):

mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
```diff
@@ -1050,15 +1050,17 @@ void mlperf_backend_clear(mlperf_backend_ptr_t backend_ptr) {
...
 
-Backend ss_backend;
+static Backend* ss_backend;
 
 // Create a new backend and return the pointer to it.
 mlperf_backend_ptr_t mlperf_backend_create(
     const char* model_path, mlperf_backend_configuration_t* configs,
     const char* native_lib_path) {
-  Backend* samsung_backend = &ss_backend;
+  ss_backend = new Backend();
```
mohitmundhragithub, line 1062: where do we clear this allocation?

al-mahi: In line#1053 `delete samsung_backend`;

android/cpp/backends/external.cc
```diff
@@ -79,7 +79,7 @@ BackendFunctions::BackendFunctions(const std::string& lib_path) {
     }
     handle = tflite::SharedLibrary::LoadLibrary(wide_lib_path.c_str());
 #else
-    handle = tflite::SharedLibrary::LoadLibrary(lib_path.c_str());
+    handle = dlopen(lib_path.c_str(), RTLD_LAZY | RTLD_LOCAL | RTLD_NODELETE);
```
sb-huh, line 82: In my opinion, this modification looks so risky, and narrow scoped.

mohitmundhragithub (approved):

mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/edenai.cc
```diff
@@ -1050,15 +1050,17 @@ void mlperf_backend_clear(mlperf_backend_ptr_t backend_ptr) {
   EdenAI::jni_CloseModel();
   EdenAI::jni_Shutdown();
   libeden_nn_on_direct = nullptr;
+  delete samsung_backend;
 }
 
-Backend ss_backend;
+static Backend* ss_backend;
```
sb-huh, line 1056: I don't understand why this declaration is needed. What is the purpose of this declaration?

Do you mean that ss_backend should hold its address of Backend for its entire life cycle?

relja128:
In the meeting we decided not to change unload behavior for all backends. Sticking with if Samsung is preferable to a wider change.

PR #198, opened 2022-01-04
Use a different main color for unofficial build.

### LINKED ISSUE #175: Flutter app: unofficial vendor builds should have a distinct skin
Allow vendors to build an unofficial version that would have a clearly different skin (brown instead of blue, "Results" -> "Unverified results").

Presence of a config file (uverified.version - can be 0 bytes) should be the key that enables this mode.

Will discuss more in the eng meeting.

anhappdev:
Closes #175.


flutter/lib/ui/main_screen.dart
```diff
@@ -133,19 +134,28 @@ class MyHomePage extends StatelessWidget {
...
+        Color(0xFF31A3E2),
+        Color(0xFF31A3E2),
+        Color(0xFF31A3E2),
+        Color(0xFF3189E2),
+        Color(0xFF0B4A7F),
+      ];
+    } else {
+      colors = [
+        Colors.brown,
+        Colors.brown,
```
d-uzlov, line 149: It think it would be better to use something like `Colors.brown.shade400` as the second color. Blank brown without any gradient doesn't look good with our UI.
Same applies to progress screen.

d-uzlov:
Do we want to keep the UI nice with this PR or is the idea to just make unofficial builds visually different, without too much concerns?

d-uzlov:
Generally I think it would be better to extract colors in some separate file. Similar to our string resources. I'm not sure if code generation would be helpful in this case but manually-written file with colors should be easier to manage than many color constants all over the code.

I don't think this should be done in this PR. This is an idea for refactoring [which may help to refine colors for unofficial builds], we should create a separate issue for it.

anhappdev:
About the color, it's just subjective, so I am open to use any color. Maybe we can discuss tomorrow in the meeting if there is any "unofficial" color which matches with the organization style we should use.

About the refactoring, yes I agree, there is also a PR #199 for this already.

d-uzlov (approved):

d-uzlov (approved):

PR #201, opened 2022-01-08
Add Markdown linter

### LINKED ISSUE #179: Add linter for Markdown
Markdown file should also have linter like other code files too.
After a quick google search, I think we can use this one:
https://github.com/avto-dev/markdown-lint

To do:
* Add doc about this linter
* Add the linter to the Dockerfile for code formatting
* Configure CI to use this linter
* Reformat all Markdown files in this repo to satisfy the linter.

anhappdev:
Closes #179 

PR #201 and #202 are related. I split into 2 separated PRs for easier review.

1. Review and merge #202 into `master` first.
2. Then merge `master` into #201, so it can pass the new Markdown linter check.
3. Review and merge #201


tools/formatter/format.mk
```diff
@@ -28,6 +28,10 @@ format/dart:
 format/line-endings:
 	git ls-files -z | xargs --null ${_start_args} dos2unix --keep-bom --
 
+.PHONY: format/markdown
+format/markdown:
+	markdownlint -c tools/formatter/markdownlint_config.yml '**/*.md' --ignore 'LICENSE.md'
```
d-uzlov, line 33: Is it possible to use `git ls-files -z` output instead of `**/*.md`?
I got an error because I had symlinks in the repo that pointed to some docker image files, and `**` could not read it.
Also `**/*.md` will include any external dependencies that are copied to the repo but added to gitignore. For example, we use it for qti backend. We don't want to format any such files.

d-uzlov (changes requested):

tools/formatter/format.mk
anhappdev: Ok. Done.

d-uzlov (approved):

Issue #209, opened 2022-01-12
Add markdown links check


d-uzlov:
We can improve our CI with this link checker: https://github.com/tcort/markdown-link-check

PR #211, opened 2022-01-12
Add markdown links check

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/209

### DIFF
--- a/tools/formatter/format.mk
+++ b/tools/formatter/format.mk
@@ -72,6 +72,10 @@ lint/line-endings:
 	else echo all files have unix line endings; \
 	fi
 
+.PHONY: lint/markdown-links
+lint/markdown-links:
+	git ls-files -z | grep --null-data "\.md$$" | xargs --null --no-run-if-empty -n1 markdown-link-check
+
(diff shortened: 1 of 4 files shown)


anhappdev (approved):

tools/formatter/format.mk
```diff
@@ -72,6 +72,10 @@ lint/line-endings:
 	else echo all files have unix line endings; \
 	fi
 
+.PHONY: lint/markdown-links
+lint/markdown-links:
```
anhappdev, line 76: Just a note. This one `lint/markdown-links` will not be called if using `docker/format`.

d-uzlov: I know. Same as with all other `lint` make targets (and some of then are not covered by the `format` target, just like this one).
I have few ideas on how to improve it, but I'm yet to formalize them to create an issue for it.

anhappdev: Ok. Because I see you added the tool to `Dockerfile` without using it, so just a note.

d-uzlov: Yeah, I launched the docker manually for this

PR #215, opened 2022-01-13
Update default task URL if changed

### LINKED ISSUE #184: Handling of updated default task file
The app uses a default task file (`tasks.pbtxt`) loaded from a URL. There are 3 cases:
1. The URL is changed, but the app is not changed.
2. The URL is changed, but the app is changed.
3. The URL is not changed, but the content of the config file is changed.

How should we handle those cases?
One idea is to use a permanent URL for the default task file. The URL will redirect to an actual task file. The actual task file should have a new name (`tasks_v1.pbtxt`, `tasks_v2.pbtxt`, ...) if the content is changed. The app will always use the latest task file.

anhappdev:
Closes #184

### DIFF
--- a/flutter/lib/resources/config_manager.dart
+++ b/flutter/lib/resources/config_manager.dart
@@ -76,13 +76,12 @@ class ConfigManager {
 
     final configs =
         jsonDecode(await file.readAsString()) as Map<String, dynamic>;
-    if (configs[defaultConfig.name] == defaultConfig.path) {
-      return file;
+    if (configs[defaultConfig.name] != defaultConfig.path) {
+      print('Update default config path in ' + file.path);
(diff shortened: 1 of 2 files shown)


flutter/lib/resources/config_manager.dart
```diff
@@ -63,12 +66,22 @@ class ConfigManager {
 
   Future<File> createConfigListFile() async {
```
d-uzlov, line 67: I think name `updateConfigListFile` would be more suitable with this change.
Also we can make this function private. The only external use for it in the `BenchmarkState.create` is useless.

anhappdev: Done

d-uzlov (approved):

Issue #203, opened 2022-01-10
Handle TFLite backend description in case of delegate fallback


anhappdev:
Per @d-uzlov's [comment](https://github.com/mlcommons/mobile_app_open/pull/194#issuecomment-1008576560):
>I should note that tflite backend uses strategy "use CPU if specified delegate failed", and we use iOS settings for Windows, so now we can see CoreML labels in the benchmark list in Windows builds.
It's not the fault of this PR, but maybe we should do something about it in the future. Probably, create an issue to create a separate config for Windows.

Possible solution: 
A `backend_description` string is passed to the backend in `RunSettings` and receive it back in `RunResult`. The TFLite backend should modify `backend_description` in case of delegate fallback.

anhappdev:
Yes. For now, I just display the backend name without the accelerator name, because of the following reasons:

1. To check which accelerator is actually used (with fallback handling) I need to modify the backend code. I want to avoid that, since the backend was well tested.
2. Sometimes TFLite can use multiples accelerators for the same model. So, it is not correct to just display one accelerator.

Therefore, weighting the time effort / side effect vs. benefits of this change, I choose to just display the backend name. This should be the same as the old Android app.

If there is a request to display accelerators name, I will add it later.

relja128:
We should display the accelerator name, since whether or not we're using CPU / GPU / NPU really matters. Let's discuss next week when everyone is around.

freedomtan:
@Mostelk and @mohitmundhragithub  please comment on if it's fine to have accelerator name displayed in QTI and Samsung' backends.

jwookiehong:
Freedom. I got you. 

@mohitmundhragithub , @Mostelk , @shabanim to investigate if providing accelerator names(DSP, GPU, CPU, NPU or any combination thereof) is possible via vendor SDKs.

shabanim:
I am confused a bit! It seems RunSettings and RunResults hold test settings. Some of these settings come from the UI such as mode. Some others are vendors specific parameters such as batch size. If my understanding is correct, device (i.e., CPU, GPU, etc.,) should be one of the parameters specified for testing.

Thus it should be specified in the generated report as well. Does this make sense to you?

freedomtan:
[...]

PR #216, opened 2022-01-13
Pass backend name from c++ to dart

anhappdev:
Part of #203

### DIFF
--- a/flutter/cpp/flutter/main.h
+++ b/flutter/cpp/flutter/main.h
@@ -38,6 +38,7 @@ struct dart_ffi_run_benchmark_out {
   char *accuracy;
   int32_t num_samples;
   float duration_ms;
+  char *backend_description;
 };
 
 struct dart_ffi_run_benchmark_out *dart_ffi_run_benchmark(
--- a/flutter/cpp/flutter/main.cc
+++ b/flutter/cpp/flutter/main.cc
@@ -91,6 +91,8 @@ extern "C" struct dart_ffi_run_benchmark_out* dart_ffi_run_benchmark(
       in->backend_native_lib_path);
   li;
 
+  char* backend_description = strdup(backend->Name().c_str());
+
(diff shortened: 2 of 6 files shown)


d-uzlov:
Are we just removing accelerator info from the results, so that we see `TFLite` instead of `TFLite | CoreML` on Windows? Is this the intention?

anhappdev:
Yes. For now, I just display the backend name, because of the following reasons:
1. To check which accelerator is actually used (with fallback handling) I need to modify the backend code. I want to avoid that, since the backend was well tested.
2. Some time TFLite use multiples accelerators for the same model. So it is not correct to just display one accelerator.

Therefore, I choose to just display the backend name. This should be the same as the old Android app.
If there is a request to display accelerators name, I will add it later.

anhappdev:
I think we can merge this PR first and leave the issue open for discussing in the next meeting.

flutter/cpp/flutter/main.h
```diff
@@ -38,6 +38,7 @@ struct dart_ffi_run_benchmark_out {
   char *accuracy;
   int32_t num_samples;
   float duration_ms;
+  char *backend_description;
```
d-uzlov, line 41: Why do we need this? I believe it would be easier to change the way we store backend name and accelerator in the `Benchmark` class (I mean, store it separately).
If anything, I would like to reduce size of the `RunResult` class. Populating `RunResult` using C++ code only increases complexity.

anhappdev: It is to pass the backend name of `ExternalBackend` from C++ to dart. 
How would you get the backend name for each task?

d-uzlov: Oh, I see. I misinterpreted this as being the same as `benchmarkSetting.configuration`.

Well, ideally I would extract the backend name from `dart_ffi_backend_match` function, but I'm not sure how much changes it would require to implement, so maybe it's not worth it.

d-uzlov (approved):

PR #217, opened 2022-01-17
[Refactor] Define color values for Flutter app in a central place

### LINKED ISSUE #199: [Refactor] Define color values for Flutter app in a central place
The color values for the Flutter app are defined inline in the code. Moving it to a central app constant would make it easier to maintain and update the app's theme.

anhappdev:
Closes #199

### DIFF
--- a/flutter/autostart-run.sh
+++ b/flutter/autostart-run.sh
@@ -25,7 +25,7 @@ submission_mode=false
 OUTPUT_LOG_PATH="output/autostart_logs"
 OUTPUT_LOG_NAME="mlperf_$(date +%Y-%m-%d-T%H-%M-%S)"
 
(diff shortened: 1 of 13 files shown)


anhappdev:
In this PR, I just move the color values to an app constant. There are quite many similar colors used. It's better to reduce the number of colors used in the app. This can be done in another PR if needed, and preferably together with UI designer.

d-uzlov:
I see there are still a lot of references to the `Colors` class outside of the `app_constants` file.
Did you intentionally skip them?

anhappdev:
Yes. Those are "local" values. It doesn't make sense to make all color values "global".

d-uzlov (approved):

Issue #219, opened 2022-01-18
Disable offline benchmark batch size manual control for Android


d-uzlov:
We have an issue related to performance in the offline benchmark: https://github.com/mlcommons/mobile_app_open/issues/159
This issue is caused by the batch size override in the Flutter app.
This override was added for the iOS version of the app, and it can probably be considered specific for the TFLite backend.
Since on Android we expect to use custom backends most of the time, and settings for the TFLite backend doesn't affect results as much as they do on iOS, the easiest way to fix performance issues on custom backends would be to just disable manual batch size management on Android.

PR #220, opened 2022-01-18
Disable batch size override on Android

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/219

### DIFF
--- a/flutter/lib/benchmark/benchmark.dart
+++ b/flutter/lib/benchmark/benchmark.dart
@@ -308,10 +308,17 @@ class BenchmarkState extends ChangeNotifier {
             }
           }
         } else if (Platform.isAndroid) {
+          // shardsCount is only supported by the TFLite backend.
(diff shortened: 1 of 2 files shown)


flutter/lib/benchmark/benchmark.dart
```diff
@@ -307,6 +307,11 @@ class BenchmarkState extends ChangeNotifier {
               break;
             }
           }
+        } else if (Platform.isAndroid) {
+          batchPreset = BatchPreset(
```
anhappdev, line 311: Can you make `shardsCount` a constant and maybe add a brief comment why is it exactly 2?

anhappdev (changes requested):

flutter/lib/benchmark/benchmark.dart
d-uzlov: Done

anhappdev (approved):

PR #222, opened 2022-01-19
Add YAML linter

### LINKED ISSUE #188: Add linter for yaml files
We don't have any checks for yaml files.
We could use [yamllint](https://github.com/adrienverge/yamllint) for local checks and [this action](https://github.com/marketplace/actions/yaml-lint) for checks on CI.

anhappdev:
Closes #188


tools/formatter/configs/yamllint.yml
```diff
@@ -0,0 +1,36 @@
+yaml-files:
+  - '*.yaml'
+  - '*.yml'
+  - '.yamllint'
+
+ignore: |
+  .github/workflows/cla.yml
```
d-uzlov, line 7: Why do we ignore this file?

anhappdev: The file `cla.yml` is used in multiple repos organization wide by the admin. If we enforce `yamllint` on this file, we need to reformat it. That makes it different from other `cla.yml`, which makes the maintenance of this file difficult for the admin.
The file `LICENSE.md` is also ignored from `markdownlint` for the same reason.

d-uzlov: Could you add this info in the file, please?

anhappdev: Done

d-uzlov (approved):

Issue #224, opened 2022-01-20
App doesn't print proper error when backend fails to load on Windows


d-uzlov:
When app can't load a backend, it calls `tflite::SharedLibrary::GetError` function.
It seems like on Windows this function doesn't work properly. When loading fails, it can return `Unknown` instead of proper error explanation, and there is no way to know what went wrong when reading the logs.

Issue #153, opened 2021-11-26
Deprecate native Android app in favor of Flutter app


anhappdev:
Since we will deprecate the Android app located in `android` folder sometime in the future. I am going to make a checklist of things to consider:

1. The Flutter app shouldn't have any dependencies on the Android app. Any shared code used by Flutter app, but located in `android` directory, should be copied to other places outside `android`.
2. Check if we can build the Android app successfully, then tag that commit and add a note to `android/README.md` that it is the last supported commit for Android app.
3. `android` folder should be marked / renamed to `android_deprecated`
4. Disable any CI steps for Android app
5. (to be updated)

relja128:
This is a good list, but we shouldn't do any of this work before we start the deprecation process (likely after the 2.0 release).

anhappdev:
We now have the Android app maintained in the `android-v2` branch, and Flutter app maintained in the `master` branch.

PR #225, opened 2022-01-20
Fix missing error on windows

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/224

### DIFF
--- a/android/cpp/backends/external.cc
+++ b/android/cpp/backends/external.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include <memory>
 #include <string>
 #include <vector>
+#include <system_error>
 
 #include "android/cpp/backend.h"
 #include "android/cpp/utils.h"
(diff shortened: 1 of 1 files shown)


anhappdev:
I don't have a Windows machine, so I cannot test this PR.

BTW, I think we should copy the whole `android/cpp` out of `android` folder to make the Flutter app independent of the Android app (this is one of the steps in https://github.com/mlcommons/mobile_app_open/issues/153) since the Android app will be deprecated soon. We have code duplicate for a short period until the Android app is deprecated, but we save us from all possible side effects when updating it for the Flutter app.

freedomtan:
@shabanim please review it.

freedomtan:
@carlosd-29 please review it

carlosd-29 (approved):
Tested and prints error message as expected.

PR #229, opened 2022-01-24
[Refactor] Refactor benchmark.dart

d-uzlov:


d-uzlov:
I moved all classes that hold info about benchmarks into the `Benchmark` class. Now it's possible to understand how we keep data without reading the whole `lib` folder, you only need to read few classes in the `benchmark.dart` file.

Next I plan to completely eliminate `BenchmarkJob` class and `BenchmarkState._getBenchmarkJobs` method.
Next it would be nice to extract actual state of the app into a separate class from the `BenchmarkState` class, which is right now an unreadable state machine.

anhappdev:
Since this PR is already quite big. I think it make more sense if you can finish all your planned refactoring in one PR then I will review and test this one big PR throughly once, instead of multiple medium PRs.

d-uzlov:
> Since this PR is already quite big. I think it make more sense if you can finish all your planned refactoring in one PR then I will review and test this one big PR throughly once, instead of multiple medium PRs.

Yeah, it's probably a good idea to finish refactoring related to the `Benchmark` in one PR.
Refactoring of `BenchmarkState`, however, should be in a separate PR in my opinion, because it would be a different kind of changes. In this PR I didn't really change anything in that class, I only moved it to a separate file without changes, and then adapted it to changes in the `Benchmark`.

I'll push more changes related to the `Benchmark` class tomorrow morning.

d-uzlov:
@anhappdev, I think I finished refactoring of everything related to the `Benchmark` class.
Please, review it when you have the time.

flutter/lib/resources/result_manager.dart
```diff
@@ -3,58 +3,31 @@ import 'dart:io';
...
+  static const _tagAccuracy = 'accuracy';
 
   final String id;
-  final double? score;
-  final String? accuracy;
-  final int shardsNum;
-  final int batchSize;
-  final String backendDescription;
+  final BenchmarkResult? performance;
+  final BenchmarkResult? accuracy;
```
anhappdev, line 14: The name `performance` and `accuracy` are cryptic when accessed outside this class. Please rename to `performanceModeResult` and `accuracyModeResult`.

flutter/lib/resources/result_manager.dart
```diff
@@ -3,58 +3,31 @@ import 'dart:io';
...
-        id: id,
-        score: score,
-        accuracy: accuracy,
-        shardsNum: shardsNum,
-        batchSize: batchSize,
-        backendDescription: backendDescription);
-  }
+      {required this.id, required this.performance, required this.accuracy});
+
+  Map<String, dynamic> toJson() => {
```
anhappdev, line 19: The name of the method is `toJson()` but return type is a `Map`. Maybe we can use `typedef Json = Map<String, dynamic>` here. The method signature would be then `Json toJson()`. And we can replace all `Map<String, dynamic>` in this file with `Json`

flutter/lib/resources/result_manager.dart
```diff
@@ -3,58 +3,31 @@ import 'dart:io';
...
+        _tagId: id,
+        _tagPerformance: performance,
+        _tagAccuracy: accuracy,
+      };
+
+  _BriefResult.fromJson(Map<String, dynamic> jsonMap)
```
anhappdev, line 25: If we use `typedef Json = Map<String, dynamic>` here. The method signature would be then `_BriefResult.fromJson(Json json)`.

anhappdev (changes requested):
Since there is a new file `device_info.dart`, the file `info.dart` is not very descriptive. I would rename the class `BenchmarkInfoItem` to `BenchmarkDescription` (since there is already a class `BenchmarkInfo`). Then rename the file `info.dart` to `benchmark_description.dart` or just move all the content in `info.dart` into `benchmark.dart`. If the `benchmark.dart` is long, we can split it into multiple files. There is a practice of "one class per file", but I am neutral to it.

flutter/lib/benchmark/benchmark.dart
```diff
@@ -84,485 +60,225 @@ class Benchmark {
...
-
-      result.add(b.taskConfig.testDataset.path);
-      result.add(b.taskConfig.testDataset.groundtruthSrc);
-
-      result.add(b.benchmarkSetting.src);
-    }
+class BenchmarkResult {
+  final double score;
+  final String accuracy;
+  final String backendName;
```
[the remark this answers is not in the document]
d-uzlov: We get backend name/description string from the function named `mlperf_backend_name`:
https://github.com/mlcommons/mobile_app_open/blob/857fe9cdd6e49528ab3df53861219a7a0ff677e6/android/cpp/c/backend_c.h#L39-L40
This function also has clear documentation that it should return name, not some other info.
I think we should rename all `backendDescription` variables in our code to `backendName` while we use `mlperf_backend_name` function to get this backend description string.

We may need to change naming if/when we use another backend function to get data.
Changing variable names in Flutter code is fast and easy, we don't depend on any vendor to do it, so I think it would be better to reflect the actual data stored in the variable.

flutter/lib/resources/result_manager.dart
d-uzlov: This is the standard naming convention for Flutter.
https://docs.flutter.dev/development/data-and-backend/json#serializing-json-inside-model-classes
I don't think using `typedef Json` would improve code.
Json is `String` of certain format.
`Map<String, dynamic>` is not json, it's one of the types that are directly encodable using standard json tools in flutter.

Maybe we could name it `JsonMap` but I doubt it would really improve readability: everyone who uses flutter [theoretically] knows standard serializable types from flutter documentation, but only people working with `result_manager.dart` will know what `JsonMap` means.

flutter/lib/benchmark/benchmark.dart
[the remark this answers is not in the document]
anhappdev: The code to read the backend name is on the Flutter side
https://github.com/mlcommons/mobile_app_open/blob/857fe9cdd6e49528ab3df53861219a7a0ff677e6/flutter/cpp/flutter/main.cc#L94
So theoretically, we can create the string `backendDescription` from multiple functions (like the not used `mlperf_backend_vendor_name` one)
Currently, the setting screen show `backend name | accelerator name` (before benchmark run) and the result screen show only `backend name`, but there is an issue to also show the `accelerator name`.
https://github.com/mlcommons/mobile_app_open/issues/203#issuecomment-1015102664
So, I prefer to just change the variable name here.

flutter/lib/resources/result_manager.dart
anhappdev: The example code is also for beginner, so they will keep it simple and don't want to explain another concept like type alias.
I think using `typedef` in this case make the code easier to read. But it doesn't impact the function of the app, so feel free to accept or ignore this change request.

flutter/lib/benchmark/benchmark.dart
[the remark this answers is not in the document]
d-uzlov: I would prefer to use naming appropriate to what the variable is _currently_ used for.
We can change the name in PR for #203. I don't know how much time it will take to specify and approve changes for #203, and then to implement it, and I would prefer to keep variable naming clear until that future PR is merged.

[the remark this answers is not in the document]
anhappdev: Ok. Also, please change `backendDescription` to `backendName` in other places, where it appropriate.

flutter/lib/ui/result_screen.dart
```diff
@@ -61,14 +62,13 @@ class _ResultScreenState extends State<ResultScreen>
...
-    final numeric = _getNumericAccuracy(accuracy);
-    switch (benchmark.type) {
+    if (benchmark.accuracy == null) return 'N/A';
+    final numeric = _getNumericAccuracy(benchmark.accuracy!.accuracy);
+    switch (benchmark.info.type) {
       // if the benchmark type is unknown, just show the original string
       // so we know that this need to be fixed
       case BenchmarkTypeEnum.unknown:
-        return _getAccuracyValue(benchmark.accuracy);
+        return _getAccuracyValue(benchmark.accuracy!.accuracy);
```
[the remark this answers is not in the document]
d-uzlov: Fixed

flutter/lib/benchmark/benchmark.dart
```diff
@@ -84,485 +60,225 @@ class Benchmark {
...
-  }
+  final BenchmarkInfo info;
+  final String backendDescription;
+  // TODO save config in the Store?
+  final BenchmarkConfig config;
 
-  Future<void> reset() async {
-    final isPreviousResultUsed = await _handlePreviousResult();
+  BenchmarkResult? performance;
+  BenchmarkResult? accuracy;
```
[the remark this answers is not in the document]
d-uzlov: Fixed

flutter/lib/benchmark/benchmark.dart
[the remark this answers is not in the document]
d-uzlov: Done

flutter/lib/benchmark/benchmark.dart
```diff
@@ -84,485 +60,225 @@ class Benchmark {
...
+  }
+}
+
+class BenchmarkList {
+  final List<Benchmark> benchmarks = <Benchmark>[];
+  final bool testMode;
+
+  BenchmarkList(String config, List<pb.BenchmarkSetting> benchmarkSettings,
+      this.testMode, BatchPreset? defaultPreset) {
+    final tasks = getMLPerfConfig(config);
```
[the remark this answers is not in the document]
d-uzlov: Done

flutter/lib/resources/result_manager.dart
d-uzlov: Done

anhappdev:
Is there a reason why the app now shows the score instead of the latency / QPS in performance mode?

flutter/lib/backend/bridge/ffi_run.dart
```diff
@@ -75,25 +76,13 @@ class _RunOut extends Struct {
   @Int32()
   external int ok;
   @Float()
-  external double latency;
+  external double score;
```
anhappdev, line 79: The name `score` is too general in this case. `latency` was more accurate, since what we calculate here is the number of query per second (so actually `1 / latency`). To be precise, we can name this `speed`, since that is what the benchmark does, measure `accuracy` and `speed`.

anhappdev (changes requested):

flutter/lib/backend/bridge/ffi_run.dart
d-uzlov: Originally we had field `latency`, but we actually transferred both latency and score, depending on the benchmark, and then we transformed it differently, depending on type of benchmark.
This is caused by loadgen. Loadgen is used by C++ code, and Flutter doesn't even know about it, so it looked weird to me  that we had logic handling weird loadgen stuff in Flutter, and I decided to move this logic to C++ code.

>To be precise, we can name this speed, since that is what the benchmark does, measure accuracy and speed.

We literary have code like `objectA.score = objectB.score` in several places in code, without modifying the value, without any logic around it.
Where would we switch from using `speed` to using `score`?
We should either use `score` everywhere, or use `speed` everywhere, or else the place where we change the name will be very strange.

`score` is used everywhere in context of different benchmarks, so I personally think we should stick with it.

anhappdev: > Originally we had field latency, but we actually transferred both latency and score, depending on the benchmark, and then we transformed it differently, depending on type of benchmark.

Can you give an example? I see only latency transferred in `cpp/flutter/main.cc`
```C++
  if (std::string{in->scenario} == "Offline") {
    out->score = driver.ComputeLatency();
  } else {
    out->score = 1000.0 / driver.ComputeLatency();
  }
```

d-uzlov: Here you go:
https://github.com/mlcommons/mobile_app_open/blob/8b16c1baa07598160d37733cd0190732bb621855/flutter/lib/backend/bridge/ffi_run.dart#L115
https://github.com/mlcommons/mobile_app_open/blob/8b16c1baa07598160d37733cd0190732bb621855/flutter/lib/benchmark/state.dart#L253
https://github.com/mlcommons/mobile_app_open/blob/8b16c1baa07598160d37733cd0190732bb621855/flutter/lib/resources/result_manager.dart#L111-L113
https://github.com/mlcommons/mobile_app_open/blob/8b16c1baa07598160d37733cd0190732bb621855/flutter/lib/ui/result_screen.dart#L75-L77
We also do this when [serializing BenchmarkResult class](https://github.com/mlcommons/mobile_app_open/blob/8b16c1baa07598160d37733cd0190732bb621855/flutter/lib/benchmark/benchmark.dart#L78), but this is irrelevant because we can change internal format for results at any time.
However, when writing `results.json` we certainly write `score` field, and I'm pretty sure everyone calls numbers on the result screen "scores". And we have `dart_ffi_run_benchmark_out.score == RunResult.score == BenchmarkResult.score`, and `RunResult.score == results.json[score]`, and `BenchmarkResult.score → number showed on result screen`.

I just don't see where we could change field name in this chain without making it confusing.

d-uzlov: Previously this logic was in the Flutter code.
This is what we currently have in the master branch:
https://github.com/mlcommons/mobile_app_open/blob/857fe9cdd6e49528ab3df53861219a7a0ff677e6/flutter/lib/benchmark/benchmark_result.dart#L51
And yeah, this is not much code, but it is still specific to loadgen.
We could technically keep the `latency` field, but I don't see a point if the result would be:
```c++
out->latency = 1000 / driver.ComputeLatency(); // if offline mode
```
```flutter
score = 1000 / out.ref.latency;
```

d-uzlov: Though, I'm fine with keeping `latency` field, it would still be easy to read.
I just don't see why it would be better than just transferring `score` field.

anhappdev: * In performance mode, the score displayed is actually `1 / latency` or query per second (QPS). This should be the same for any benchmarks.
* In accuracy mode, the score displayed is the accuracy of each benchmark. This will be different for each benchmark, since each benchmark has a unique method to calculate accuracy.
So I think `speed` would be more accurate than `score`, or we can keep the `latency` name.

I guest the confusion with the 1000 is a legacy thing. It looks like the app sometimes showed `latency in ms` or `qps` (= `1000 / latency in ms`). But now they all show in `qps`, that is why the conversion are all over the place.
https://github.com/mlcommons/mobile_app_open/blob/857fe9cdd6e49528ab3df53861219a7a0ff677e6/android/cpp/mlperf_driver.h#L78-L90

If we want to fix it, then we should change it here and make this function return only `qps`
https://github.com/mlcommons/mobile_app_open/blob/857fe9cdd6e49528ab3df53861219a7a0ff677e6/android/cpp/mlperf_driver.h#L66-L75

d-uzlov: I see.

Changes in mlperf driver would affect the old Android app.
I do remember that we have an [issue](https://github.com/mlcommons/mobile_app_open/issues/153) to deprecate the old app, including separating its C++ code from Flutter, but Relja has explicitly asked us to delay it.

I suggest right now we just rename field `score` to `qps`, and then maybe do C++ code refactoring after old app deprecation.

anhappdev: `qps` is a unit of measurement similar like `ms`. So I prefer to name the variable `speed` or `latency`.

d-uzlov: Ok, I changed qps to latency

anhappdev: I see that `score` is still used instead of `latency` in many places, e.g. in `BenchmarkResult`, `RunResult`.

d-uzlov: Because it it was always this way. I only renamed that field in C++ struct, so I reverted this change. In all other places I just moved what we had to different files, but I kept the original naming.
We also use `score` in tasks file:
https://github.com/mlcommons/mobile_models/blob/dec215a473e78e52d0d21806a8f24875bb50aca4/v1_0/assets/tasks_v3.pbtxt#L26
We also use `score`+`accuracy` in `result.json`:
https://github.com/mlcommons/mobile_app_open/blob/455e69fcbea2474ac238b2223ee32f7e652b617c/flutter/lib/resources/result_manager.dart#L79-L80
I believe that field names in `result.json` must never change (or at least an approval is required).

d-uzlov: I think we should discuss changing the name in a separate issue.

anhappdev: Ok.

anhappdev (approved):

PR #234, opened 2022-01-27
Patch the latest NMS implementation from TF master

ptt2panda:
@freedomtan 
Could we add this patch to pixel's backend?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -12,10 +12,7 @@ http_archive(
 http_archive(
     name = "org_tensorflow",
     patch_args = ["-p1"],
-    patches = [
-        # Fix tensorflow not being able to read image files on Windows
-        "//:flutter/third_party/tensorflow-fix-file-opening-mode-for-Windows.patch",
-    ],
+    patches = ["//patches:fast_nms.diff"],
     sha256 = "40d3203ab5f246d83bae328288a24209a2b85794f1b3e2cd0329458d8e7c1985",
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -12,10 +12,7 @@ http_archive(
 http_archive(
     name = "org_tensorflow",
     patch_args = ["-p1"],
-    patches = [
-        # Fix tensorflow not being able to read image files on Windows
-        "//:flutter/third_party/tensorflow-fix-file-opening-mode-for-Windows.patch",
```
d-uzlov, line 17: This patch is still required for the Flutter app

d-uzlov (changes requested):

WORKSPACE
ptt2panda: Fixed, thanks.

freedomtan: Yes, please keep this windows patch.

And maybe make this a 1.1 only patch so that we don't have to remove it when TF 2.7 or later is used to replace 2.6.

ptt2panda: How do I make it 1.1 only?

freedomtan: @d-uzlov and @anhappdev please create a new 1.1 branch (supposedly based on the branch you used to created the intergrated 1.1 apk)

anhappdev: I created a new branch 'release-1.1-Jan-2022'

PR #245, opened 2022-02-08
Rename score to throughput

### LINKED ISSUE #237: Inconsistency in naming variable for speed measurement
Currently, in the app, we have different variable names to mean the same thing:
* `score` measured in `QPS` (Query Per Second)
* `latency` measured in `ms` for single stream benchmark
* `latency` measured in `QPS` for offline benchmark
* `score` is also used to measure accuracy with different units depend on benchmark type.

This caused confusion and conversion between units (`qps = 1000 / latency in ms`) in many places.

We should agree on one name and unit for speed measurement to be consistent.

anhappdev:
Closes #237

### DIFF
--- a/flutter/lib/resources/result_manager.dart
+++ b/flutter/lib/resources/result_manager.dart
@@ -43,7 +43,7 @@ class RunInfo {
 
 class _FullResult {
   final String id;
-  final String score;
+  final String throughput;
   final String accuracy;
   // numeric values are saved as string to match old android app behavior
   final String minDuration;
@@ -58,7 +58,7 @@ class _FullResult {
 
   _FullResult(
       {required this.id,
-      required this.score,
+      required this.throughput,
       required this.accuracy,
       required this.minDuration,
       required this.duration,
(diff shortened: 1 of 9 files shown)


flutter/lib/resources/result_manager.dart
```diff
@@ -76,7 +76,7 @@ class _FullResult {
       'configuration': {
         'runtime': '',
       },
-      'score': score,
+      'throughput': throughput,
```
d-uzlov, line 79: Are we allowed to change format of `result.json`? I don't think we discussed it in the last meeting.

I remember that I was requested to compare formats of `result.json` generated by the old app and by the Flutter app.
We still have some strange things just to make `result.json` fully equivalent to the old app result file format. For example, saving numeric values in strings, and manually adding `\` to `/`, as if it is a special character.

d-uzlov: I created an issue to discuss it: https://github.com/mlcommons/mobile_app_open/issues/247

d-uzlov: We can probably ignore it for now, and revert this field name change later if `result.json` must have `score` field.

d-uzlov (approved):

PR #250, opened 2022-02-16
Decoupling android and flutter app

### LINKED ISSUE #249: Decoupling Android and Flutter app
The Android and Flutter app are currently depended on the same module `android/cpp`.
There are some issues (Flutter app related) which require changes in `android/cpp`. Since we are going to deprecate Android in the next version, I suggest we **copy** `android/cpp` into `flutter/cpp` so that the changes don't impact the Android app.

anhappdev:
Part of #249


d-uzlov:
I'm really not sure if this actually helps us.

We wanted to separate flutter and android `cpp` folders to make sure that changes we make for the Flutter app do not affect the android app.
But after this PR all backends will point to `flutter/cpp`. Which means that if we make any changes in the `flutter/cpp`, and then try to build the Android app, we will build it with potentially incompatible backends. Even if we don't include any vendor backends, we will still build the tflite backend, which also points to `flutter/cpp`.
This completely defeats the purpose of having 2 separate `cpp` folders.

I think the only way to keep the Android app stable while making changes for the Flutter app is to also make a copy of all the backend folders.
And this sound like quite a bit of work to make sure that backend sources for Android app and for Flutter app don't reference each other, and each app only references its dedicated backends. Not to mention increased clutter in the repo.

I think there are only 2 viable ways here:
* Deprecate Android app right now, move `cpp` folder and remove `android` folder from the master branch
  * We could still develop it in other branches
  * This is basically "make copies of all backends" solution, we would just use Git to manage copies
* Keep `cpp` in one place, and check that old app still works
  * Like what we have been doing for that last few months

I personally would choose the second solution, unless we really need breaking changes that would be difficult to adapt for the Android app.

As far as I know currently the only breaking change we want to do is adding `mlperf_backend_accelerator_name` function to the backend interface, and this should not require much time to adapt the Android app to it. In fact, I think we won't need any changes there at all.

freedomtan:
yup, I am more with @d-uzlov. Moving android/cpp/ to cpp/  and changing both android/ and flutter/ to refer to it should work.

anhappdev:
I would also do the second one (both Android and Flutter still share the same `cpp` module) if there is only adding `mlperf_backend_accelerator_name`, but there is also other change which requires updating the Android app,
see https://github.com/mlcommons/mobile_app_open/blob/149b224fd5b19179bc6acd5418a243e4d7e89372/flutter/cpp/flutter/main.cc#L145-L148

Probably we will have other changes in the future which impact both Android and Flutter, so it's only the question **when** should we decouple it.

So I think the first solution would be more future-proof, we use the master branch for Flutter app, and have another branch (maybe the `submission-v2.0`) to maintain the Android app.

d-uzlov:
Let's discuss splitting the Android app to a separate branch on the next weekly meeting.

freedomtan:
To be future-proof, how about
1. a separate directory, and
2. versioning the interface for both frontend/ui and backend. e.g. for backend, we can add a version number field or function (something like `int32_t mlperf_backend_version(void)`)"

d-uzlov (approved):

freedomtan:
Let me do a quick check within the next two hours.

freedomtan (approved):

Issue #253, opened 2022-02-16
Add result history to the Flutter app


d-uzlov:
Currently the app only shows last result. When you rerun any set of benchmarks, results of the previous run are erased.
We need to add a new UI page that will show list of previous results and allow to inspect them.

This feature must be offline.
We can technically upload result into Firebase, and then fetch them to show in history but sometimes this can be undesirable, so we must keep results locally.

Initial discussion was here: https://github.com/mlcommons/mobile_app_open/issues/235

Steps:
* [x] Add history saving
* [ ] Add UI pages for result list and for individual results

freedomtan:
@anhappdev please help on this one. do some initial investigation

anhappdev:
I would store the `result.json` file of each benchmark run on disk with a timestamp. In the history screen, the app shows a list of those result files, and on tap of one result file, the app show the content of the result file.

d-uzlov:
When working on https://github.com/mlcommons/mobile_app_open/issues/251 I have modified the way we handle `result.json` a bit. I fixed serialization and added few new fields that are required for uploading.
I'll extract it into a separate PR today so you can use it for result history.

anhappdev:
@d-uzlov Would you take over this issue? I think with the tasks you are current doing (implementation of new result spec), this issue would fit better in your workflow and design.

d-uzlov:
@anhappdev sure.

anhappdev:
@d-uzlov Great. Thanks.

d-uzlov:
I have created 2 separate issues for backend changes and for UI changes.

Issue #251, opened 2022-02-16
Add basic result uploading


d-uzlov:
Steps:
1. Create Firebase Cloud Firestore database
2. Create REST API service Firebase Cloud Functions that will provide basic access to this database
    * We only need 2 methods for now: create new result and get all results
    * We will create separate github issues to implement more functions once this task is completed
3. Add Firebase anonymous authentication to the Flutter app (plugin [firebase_dart](https://pub.dev/packages/firebase_dart) can be used for this)
4. Connect the Flutter app to this REST service

Initial discussion was here: https://github.com/mlcommons/mobile_app_open/issues/235

PR #259, opened 2022-02-25
Add proper full result serialization

d-uzlov:
This PR adds functions to de/serialize full results, as well as additional information that is needed for result uploading and history.

Part of https://github.com/mlcommons/mobile_app_open/issues/251
Related to https://github.com/mlcommons/mobile_app_open/issues/253


flutter/lib/data/extended_result.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:mlperfbench/data/export_result.dart';
+import 'environment_info.dart';
+
+class ExtendedResult {
+  static const String _tagUuid = 'uuid';
+  static const String _tagResultJson = 'results';
+  static const String _tagEnvInfo = 'appVersion';
```
anhappdev, line 7: I would use the same string here, either app version or env info.

flutter/lib/data/export_result.dart
```diff
@@ -0,0 +1,123 @@
...
+        _tagNumSamples: numSamples,
+        _tagSharsNum: shardsNum,
+        _tagBatchSize: batchSize,
+        _tagMode: mode,
+        _tagDatetime: datetime,
+        _tagBackendName: backendName,
+      };
+}
+
+class ExportResultList {
```
anhappdev, line 103: I think it quite common to use singular noun for one object and plural noun for a list of objects, so `ExportResults` should be sufficient.

anhappdev (changes requested):

flutter/lib/data/export_result.dart
d-uzlov: I wasn't able to find too much info on best practices of naming collections.
One article I found is here: https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098
It looks very reasonable to me.
It suggests always using singular nouns for class names. Plural nouns are fine for variable names but in class names they are harder to understand.

anhappdev: Naming convention is more or less subjective, so I am fine with any style if it's consistent. We already have some class with plural noun:
- Icons
- AppColors
- RunSettings
- ResultKeys
- ShareKeys
- ..

d-uzlov: The difference between `ExportResults` and other plural class names in the project is the existence of `ExportResult` class.
We don't have a `ShareKey`, `RunSetting`, or `AppColor` class. If we had any of those, I would argue that having both singular and plural class name is confusing.
Also, `Icons` is a bad name in my opinion, both because it collides with the standard library name and because there is an `Icon` class.

anhappdev: Yes. We need some more refactoring to make the naming consistent.

anhappdev (approved):

Issue #252, opened 2022-02-16
Add a website to observe uploaded results


d-uzlov:
We need a way to access uploaded results via a browser.
We can use Firebase Hosting for static content.

A relatively easy way to create a website would be to use Flutter Web.
Flutter Web has some issues, it does not provide native web experience, but we can use it to create a prototype faster, and then switch to something else later, if we need it.

Depends on: https://github.com/mlcommons/mobile_app_open/issues/251

Initial discussion was here: https://github.com/mlcommons/mobile_app_open/issues/235

Status:
* [x] Create Flutter project for the website
* [x] Create Flutter project for shared code (shared between app and website)
* [x] Extract shared code from app project to shared project
* [x] Add simple pages to show dummy results data
* [x] Add `fetch` method to our firebase web service
* [x] Connect website to firebase (show real data from database)
* [x] Modify `fetch` method of the web service to use pagination
* [x] Add pagination support to website
* [x] Improve website appearance for easier reading
* [x] Include shared and website Flutter projects to linter checks
* [x] Deploy website in firebase
* [x] Add documentation for website deployment and maintenance

d-uzlov:
Right now I have a simple website prototype written in flutter but our firebase service does not support fetching results, so it can only show dummy data.
I'm working on connecting the website to firebase.

Issue #247, opened 2022-02-09
Formal specification for result.json file


d-uzlov:
Currently I don't see any info about the structure of this file.
As far as I'm aware it is imported in some external tools, so it must have a certain format which is recognized by these tools.

We need to write a proper specification for this file, so we can stick to it when making changes in sources, to avoid breaking any external tools.

freedomtan:
There should be some descriptions in the submission rule doc. Let's if what we can do from that doc.
@freedomtan to add submission rules doc link later

d-uzlov:
In the old Android app there is a function that creates `result.json`:
https://github.com/mlcommons/mobile_app_open/blob/084729b7f29043dc833fd7a90c49940fc5012e3e/android/java/org/mlperf/inference/MLPerfTasks.java#L149

## Android app

It creates the following structure:
On top level there is an array of benchmark entries. Each entry is a map with the following fields:
```json
{
"benchmark_id": string,
"configuration": {
  "runtime": string
},
"score": string,
"accuracy": string,
"min_duration": string,
"duration": string,
"min_samples": string,
"num_samples": string,
"mode": string,
"datetime": string
}
```
Few notes about these fields:
* `runtime` is always empty. In the app sources I see that it was supposed to always be `External` instead of empty string, but in the real results file I got after running the app it's just empty.
Both empty and "External" values doesn't make much sense to me, if they never change.
* `score` always contains benchmark score
* `accuracy` for contains accuracy string that we get from native code for submission mode results, and `N\/A` for performance mode results.
I don't understand why there is `\/` instead of just `/`. It looks as if `/` was a special JSON character which must be escaped, but it is not.
* `score`, `min_duration`, `duration`, `min_samples`, `num_samples` are numeric values but we write them as strings for some unknown reason.
* `datetime` has the following form: `2022-02-15T15:04:52.291`. It is missing time zone.
* `mode` can have one of the following values:
  * `performance_lite_mode`
  * `accuracy_mode`
  * It seems like at some point other run modes were used, but now the app only runs these 2 modes.
* `min_duration` and `min_samples` values are from benchmark config
* `duration` and `num_samples` are measured values

## Flutter app

Flutter app manages result file creation in this class:
https://github.com/mlcommons/mobile_app_open/blob/084729b7f29043dc833fd7a90c49940fc5012e3e/flutter/lib/resources/result_manager.dart#L44
Flutter app mimics Android app result file, it has the same fields, with the same type of values.
[...]

PR #260, opened 2022-03-01
Add result uploading

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/251


d-uzlov:
This PR adds only upload function. The only way to see the list of previously uploaded results is through the Firestore console.
Later we will add the ability to observe results in a web browser in https://github.com/mlcommons/mobile_app_open/issues/252.
We could also reuse changes from https://github.com/mlcommons/mobile_app_open/issues/253 to show results in the app when this issue is done.

d-uzlov:
Data structure in database is subject to change.
I haven't actually deployed Firebase Functions for our main Firebase project yet, I have used my personal project for this when testing, so there will be no issues due to structure changes.

Amount of data that we save is also subject to change.
We have a separate issue about result structure and what to save: https://github.com/mlcommons/mobile_app_open/issues/247
In this PR I'm saving everything as-is.

flutter/lib/firebase/rest_helper.dart
```diff
@@ -0,0 +1,29 @@
...
+
+  RestHelper(this.userInfo);
+
+  Future<void> upload(ExtendedResult result) async {
+    const path = '/v0/upload';
+    final jsonResult = result.toJson();
+    var response =
+        await http.post(Uri.parse('${FirebaseConfig.functionsUrl}$path'),
+            headers: {
+              'Authorization': '${await userInfo.user!.getIdToken()}',
```
anhappdev, line 21: Force unwrap of user may cause crash.

flutter/lib/benchmark/state.dart
```diff
@@ -84,11 +85,15 @@ class BenchmarkState extends ChangeNotifier {
 
   late BenchmarkList _middle;
 
-  BenchmarkState._(this._store, this.backendBridge) {
+  BenchmarkState._(this._store, this.backendBridge, this.firebaseManager) {
     resourceManager = ResourceManager(notifyListeners);
     backendInfo = BackendInfo.findMatching();
   }
 
+  Future<void> testUpload() async {
```
anhappdev, line 93: What does “testUpload” mean?

flutter/lib/main.dart
```diff
@@ -32,8 +33,10 @@ Future<void> main() async {
 
 Future<void> launchUi() async {
   await initDeviceInfo();
+  await FirebaseManager.staticInit();
+  final firebaseManager = await FirebaseManager.create();
```
anhappdev, line 37: Must `FirebaseManager.staticInit()` always be called before `FirebaseManager.create()`? Can we integrate `FirebaseManager.staticInit()` into `FirebaseManager.create()`?

anhappdev (changes requested):

flutter/lib/main.dart
d-uzlov: `FirebaseManager.staticInit` must be called exactly once. It will throw an error if called for a second time.
In theory it's possible to create several instances, so I don't think we should integrate `staticInit` into `create` method. It's just a coincidence that we call both functions in one place.
That being said, `create` in its current implementation will also throw an error if called several times, so this must be fixed. I'll make it a singleton to reflect how underlying library works.

flutter/lib/benchmark/state.dart
d-uzlov: I created this function when I first connected the app to firebase and it didn't function properly, and forgot to rename it when I correctly implemented the upload.
Fixed.

flutter/lib/firebase/rest_helper.dart
d-uzlov: Fixed, I added a separate check.
Though, this function is expected to throw errors, so calls to it were wrapped into try-catch block from the start. Explicit check just makes the error more readable.

anhappdev: I did not test it but I think force unwrap will crash the app and user has no idea what happened, but with try-catch we can show a popup with an error message.

anhappdev (approved):

PR #261, opened 2022-03-04
Disable TFLite as default backend when building app.

### LINKED ISSUE #205: Backend hardware match is not handled as expected in flutter app
Found that in the latest flutter based app, if qti backend detects a snapdragon device, but the device is not supported, then also, the app goes ahead and loads the next backend and finally runs the tflite models.
Expected: The app should show a message saying that the device is unsupported instead of loading the next backend. This is what happens in the legacy android app, but not in the flutter based android app.

anhappdev:
Closes #205

### DIFF
--- a/.github/cloudbuild/flutter-android.yaml
+++ b/.github/cloudbuild/flutter-android.yaml
@@ -69,7 +69,7 @@ steps:
     args:
       - -xc
       - |
-        make flutter/android/test_apk || exit $?
+        make flutter/android/test_apk WITH_TFLITE=1 || exit $?
         mkdir -p /workspace/output/artifacts || exit $?
         cp /workspace/flutter/build/app/outputs/flutter-apk/app-release.apk /workspace/output/artifacts/mlperf_app_release.apk || exit $?
         cp /workspace/flutter/build/app/outputs/apk/debug/app-debug.apk /workspace/output/artifacts/mlperf_app_test_main.apk || exit $?
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,14 @@
 
 all: android/app
 
-# avaiable backends
-WITH_TFLITE?=1
+# Available backends. App will crash if phone and backend do not match.
 WITH_QTI?=0
 WITH_SAMSUNG?=0
 WITH_PIXEL?=0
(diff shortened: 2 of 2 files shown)


freedomtan:
No, no, as far as I can remember and what I expect is that when we run it on and unsupported Qualcomm SoC, it will recognize it's Q's platform but not supported. So something like "hey, we know this is a Qualcomm SoC, unfortunately, we don't support it yet"

anhappdev:
1. Do you have a screenshot from the old Android app to see how the UI would look?
2. What should happen if the app is built with multiple backends and TFLite is one of it? Should it show pop up for every backend tried but not found, then run with TFLite as the last one?

freedomtan:
FYR. This is what 1.1 Feb 1st release apk on a Motorola Edge X30 shows.

freedomtan:
> 2. What should happen if the app is built with multiple backends and TFLite is one of it? Should it show pop up for every backend tried but not found, then run with TFLite as the last one?

Nothing pops up. It just goes to TFLite + NNAPI.

anhappdev:
Ok. Thanks. I will update the UI accordingly.

d-uzlov:
@anhappdev see here:
https://github.com/mlcommons/mobile_app_open/blob/40e01139b89e73acada30f0966f7d116ad7cdc7b/mobile_back_qti/cpp/backend_qti/qti_c.cc#L72
We must handle this error message to determine if we can skip current backend or if we must stop iterating over backends.
For Flutter app we do checks here:
https://github.com/mlcommons/mobile_app_open/blob/40e01139b89e73acada30f0966f7d116ad7cdc7b/flutter/cpp/flutter/dart_backend_match.cc#L38
Currently we only log this error, we need to modify this `dart_ffi_backend_match` interface to support stopping at a backend.

anhappdev:
@d-uzlov Can you take over this issue? I think it would be more productive since you have worked from the beginning with that code part.

d-uzlov:
@anhappdev ok, I'll work on this when I finish pull request for result uploading (hopefully today).

anhappdev:
Thank you.

PR #262, opened 2022-03-10
Remove `android` folder from main branch

### LINKED ISSUE #249: Decoupling Android and Flutter app

anhappdev:
Closes #249 

This PR removes `android` folder from the main branch.
From now on, Flutter app will be maintained on `master` branch and legacy native Android app will be maintained on `android-v2` branch.


anhappdev:
I split the changes in multiple commits, so it should be easier to review by looking at each commit at a time.

d-uzlov (approved):

PR #265, opened 2022-03-14
Add new API to return the accelerator name.

### LINKED ISSUE #203: Handle TFLite backend description in case of delegate fallback

anhappdev:
Closes #203.

Each vendor should modify the function `mlperf_backend_accelerator_name` and return the appropriate accelerator name.

### DIFF
--- a/flutter/cpp/backend.h
+++ b/flutter/cpp/backend.h
@@ -38,6 +38,9 @@ class Backend {
   // A human-readable string for logging purposes.
   virtual const std::string& Name() const = 0;
 
+  // Accelerator name.
+  virtual const std::string& AcceleratorName() const = 0;
+
   // Run inference for a sample. Inputs is already set by SetInputs.
   virtual void IssueQuery() = 0;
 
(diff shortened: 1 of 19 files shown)


anhappdev:
For TFLite backend, @anhappdev or @d-uzlov will implement the method in another PR.
For other backends, each vendor should submit a separated PR.

freedomtan (approved):
LGTM

d-uzlov (approved):

PR #267, opened 2022-03-15
Add specification for results

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/247


freedomtan:
let's try to finish reviewing this by next meeting

flutter/documentation/result-spec.md
```diff
@@ -0,0 +1,129 @@
...
+  Extra settings that can vary between different benchmarks and backends.  
+  Here must be stored values set by backend in `common_setting`.  
+  `batch_size` value for offline benchmarks should be located here.  
+  `shards_num` value for TFLite backend should be located here.  
+  Map structure:
+  * `id`: string. Value from `setting.id` that is passed to backend
+  * `name`: string. Value from `setting.name` that is passed to backend
+  * `value`: string. Value from `setting.value.value` that is passed to backend
+  * `value_name`: string. Value from `setting.value.name` that is passed to backend
+* `throughput`: floating point number. Present for all results
```
freedomtan, line 69: question for all. Should we specify precision (number of digits after the decimal point)

flutter/documentation/result-spec.md
```diff
@@ -0,0 +1,129 @@
...
+* `accuracy`: floating point number. Present for all results, including `loadgen_mode=PerformanceOnly` runs  
+  When actual accuracy is not available value `-1` must be used
```
freedomtan, line 71: Does JSON support [NaN](https://en.wikipedia.org/wiki/NaN)?

flutter/documentation/result-spec.md
```diff
@@ -0,0 +1,129 @@
...
+* `official_release_flag`: bool  
+  Indicates if the official release flag was set for this build
+* `dev_test_flag`: bool  
+  Indicated if development test flag was set for this build
+* `backend_list`: list of strings  
+  Must contain actual list of backends that are included into this version of the app.
+* `git_branch`: string
+* `git_commit`: string
+* `git_dirty_flag`: bool  
+  Indicates if there are any local changes compared to the commit specified in `git_commit`.
```
freedomtan, line 129: These build info look good to me.

question for all, should we have backend specific version info like these (e.g., SNPE version, Samsung ENN SDK version, and MTK Neural Delegate version)

flutter/documentation/result-spec.md
d-uzlov: I think precision must be part of UI. It would be strange if it could change between different results.

flutter/documentation/result-spec.md
d-uzlov: NaN is not part of JSON specification.
It could still theoretically be supported by the tools/libs we use as an extension to JSON but it's not guaranteed.

flutter/documentation/result-spec.md
d-uzlov: We would need to get this info from somewhere in the build time.
SNPE version can be obtained from SNPE SDK folder name, but I don't know where we could automatically get versions for other backends.

jwookiehong:
Errrr.... looks like I'm not a reviewer. @AhmedTElthakeb, could we go over this?

flutter/documentation/result-spec.md
mohitmundhragithub: We have the flexibility to use SNPE or TFLite for different models run in a single execution. So, if we need to, we may have to add seperate delegate names for separate use case. e.g., on SDM865, we use TFLite GPU delegate for MobileBERT and SNPE for rest of the models.

mohitmundhragithub (approved):

flutter/documentation/result-spec.md
d-uzlov: >we may have to add seperate delegate names for separate use case

This should be in benchmark-specific fields, not in build info.

We have `backend_accelerator_name` field there. I think it should suffice for this.

PR #272, opened 2022-03-28
Add website

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/252

Demo of this PR is online here: https://mlperf-dev-test-2.web.app/#/


freedomtan:
@jwookiehong to bring this web site prototype to the general to get more opinions

website/lib/main.dart
```diff
@@ -0,0 +1,69 @@
...
+  await FirebaseManager.staticInit();
+}
+
+class MyApp extends StatelessWidget {
+  const MyApp({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      title: 'Flutter Demo',
```
anhappdev, line 35: Title can be updated.

anhappdev (commented):
I am not sure if we should merge this change into master right now. Maybe we can keep it in a separated branch.
The separated can be in sync with master by regularly merging master into it.

d-uzlov:
@anhappdev
Which part of these changes do you think we should keep in a separate branch?

As I see it there are 3 relatively independent parts:
1. Creation of the `flutter_common` Flutter project
2. Changes in firebase service
3. Creation of the `website` Flutter project

These changes don't make sense on their own.
Website wouldn't work without other 2 changes, new firebase methods can't be tested without website (at least until we implement result browsing in the mobile app), `flutter_common` is not needed if we don't add the website project.

We could keep the website folder in a separate branch but if we don't merge flutter_common and firebase changes it would be hard to merge them later.

anhappdev:
@d-uzlov The whole PR. The website is very basis right now and needs further updates. But as I understand (correct me if I am wrong) the website doesn't have high priority at the moment. If we merge it now the whole app is more complex, more difficult to maintain.
So in my opinion I would develop the website in a different branch until it's more matured, and then merge it into master.

d-uzlov:
@anhappdev 
Which part of the PR makes the app more complex? Maybe it would be better to fix it instead of maintaining a separate branch.

anhappdev:
@d-uzlov Not any certain part but the idea of integrating a website into the app. But it's just my opinion. Maybe we can discuss it in the meeting.

d-uzlov:
@anhappdev 
Are you concerned about shared code between the app and the website, or by Flutter Web overall?

Shared `flutter_common` project doesn't bring any major structure changes. We already had firebase-related dart files in separate folders inside `flutter/lib`. This PR moves them to `flutter_common/lib`.
It also doesn't bring any issues for debugging. VSCode allows you to edit and debug both projects as if it was one project.
If anything, this simplifies things because we only need to maintain one set of files to manipulate Firebase, so we can't have a mismatch between the app and the website.

We could potentially decide get rid of Flutter Web for the website and reimplement it on some other framework.
But this would only mean that we will have to delete the `website` folder and move `flutter_common/lib` back to `flutter/lib`.
Changes from this PR related to firebase would still remain. We would still need them to implement result browsing in the app.

d-uzlov:
I could extract firebase-related changes to a separate PR, leaving website and `flutter_common` folder in a separate branch if you consider them harmful, but we will not be able to fully test these changes.

anhappdev:
Ok. Sounds good to me.

anhappdev (approved):

Issue #274, opened 2022-04-01
Submission mode doesn't run accuracy benchmark


d-uzlov:
When submission mode switch in settings is enabled calculated accuracy is `0.0`, and logs indicate that both performance and accuracy benchmarks are actually running in performance only mode.

PR #275, opened 2022-04-01
Fix accuracy mode

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/274

### DIFF
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -274,7 +274,9 @@ class BenchmarkState extends ChangeNotifier {
       benchmark.accuracyModeResult = BenchmarkResult(
           // loadgen doesn't calculate ltency for accuracy mode benchmarks
           // so throughput is infinity which is not a valid JSON numeric value
-          throughput: 0.0,
+          throughput: accuracyResult.result.throughput.isFinite
+              ? accuracyResult.result.throughput
+              : 0.0,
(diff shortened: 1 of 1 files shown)


d-uzlov:
While investigating this I tried using the test dataset but it didn't work.
In our current tasks file test datasets has groundtruth files specified so I thought it would be suitable for accuracy mode testing. But apparently those groungtruth files don't match inputs of test datasets so accuracy is zero when using them.

flutter/lib/benchmark/state.dart
```diff
@@ -266,13 +266,15 @@ class BenchmarkState extends ChangeNotifier {
...
+      final accuracyResult = await runBenchmark(benchmark, true,
           backendInfo.settings.commonSetting, backendInfo.libPath);
       _updateProgress(doneCounter * doneMultiplier / activeBenchmarks.length);
       doneCounter++;
 
       benchmark.accuracyModeResult = BenchmarkResult(
-          throughput: accuracyResult.result.throughput,
+          // loadgen doesn't calculate ltency for accuracy mode benchmarks
+          // so throughput is infinity which is not a valid JSON numeric value
+          throughput: 0.0,
```
anhappdev, line 277: A condition check would be better in case `accuracyResult.result.throughput` is valid in the feature.

d-uzlov: Done

anhappdev (approved):

Issue #278, opened 2022-04-05
Merge changes from v2.0 submission


freedomtan:
we have at least the following issues
1. the latency setting,
2. the TensorFlow 2.80, 
3. uploading new (converted, e.g., QIT's dlc files and original (mosaic)) 

Let's start the merging after the publication this Wed (Apr 6th)

freedomtan:
@anhappdev   please merge  https://github.com/mlcommons/mobile_app_open/tree/submission-v2.0 to both the Android branch and main/master branch.

freedomtan:
@fatihcakirs has submitted some nnc files. @anhappdev to help put all the .nnc and .dlc files to the mobile_models repo

anhappdev:
Trying upload model files for Qualcomm I got an error with size limit:
```
remote: error: File v2_0/SNPE/mobilebert_quantized_htp.dlc is 104.35 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
```

Should we use Git LFS for model files?

freedomtan:
yeah, large file is an issue. As far as I can remember, github lfs has monthly bandwidth limit (1 GiB per month). https://docs.github.com/en/enterprise-cloud@latest/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage

That is, after downloading > 10 times, we'll run out of 1 GB / month quota :-(

anhappdev:
Is there other alternative storage provider we can use?

>Technically, we can work around this by compressing or splitting large files. We need way to decompress the compressed file or merge splitted files.

Yes, it's possible but I am not sure if it's worth the effort to implement it vs. just using a storage service.

freedomtan:
@willc2010 and @jwookiehong what's your opinion?

Mostelk:
If possible to gauge how much work is needed for this issue, as it will be needed for future submission as well https://github.com/mlcommons/mobile_app_open/issues/291

freedomtan:
Possible solutions

1. buy storage services (github or other version control services), 
2. splitting (and/or or compressing)
3. ?

freedomtan:
@jwookiehong will check with @TheKanter to see if we can pay for services

jwookiehong:
I talked to David, and he is receptive to the idea. Now we need to figure
out which services we need, and how much it will cost. Also, is it possible
[...]

PR #286, opened 2022-04-13
Merge submission-v2.0 into android-v2

anhappdev:
Part of #278


anhappdev:
The  failed test`Build and test iOS Flutter app` will be deleted in the `android-v2` branch, so it can be ignored.

anhappdev:
The current `submission-v2.0` branch do not have the changes from the vendors. I need to apply these first.

PR #292, opened 2022-04-14
Merge submission v2.0 into master

anhappdev:
Part of #278

PR #297, opened 2022-04-28
Update config for Markdown links check

### LINKED ISSUE #296: Reconfigure Markdown links check
The Markdown links check failed on 403 forbidden status (for example this link: https://help.github.com/articles/using-pull-requests/) but the browser can open and redirect the link.

anhappdev:
Closes #296

### DIFF
--- a/.github/workflows/format-and-lint.yml
+++ b/.github/workflows/format-and-lint.yml
@@ -85,7 +85,6 @@ jobs:
       - name: Check links in markdown files
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
-          config-file: 'tools/formatter/configs/markdownlinkscheck.json'
           use-verbose-mode: yes
           use-quiet-mode: yes
   yaml:
--- a/.github/workflows/format-and-lint.yml
+++ b/.github/workflows/format-and-lint.yml
@@ -85,7 +85,9 @@ jobs:
       - name: Check links in markdown files
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
+          config-file: 'tools/formatter/configs/markdownlinkscheck.json'
           use-verbose-mode: yes
+          use-quiet-mode: yes
   yaml:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


d-uzlov:
Generally I don't like the idea of treating 4xx codes as success codes. I guess 403 could mean that CI just doesn't have access to some resource but it could also mean that resource is no longer available and we should fix the link.

I intentionally didn't mark `Format & lint / markdown` check as a requirement for pull requests because I expected some fails which may not have anything with the specific pull request.

Where is that failed check that you mentioned in the https://github.com/mlcommons/mobile_app_open/issues/296?
Is it permanent, or is it happening only sometimes?

If this specific link often (or always) fails to load on CI, I think it would be better to make an exception for this specific link.
Previously I excluded some links in the datasets directory from the check because they are located in the private repository which is not accessible from the CI.
Se here: [datasets/util/ade20k/README.md](https://github.com/mlcommons/mobile_app_open/blob/4f72934ca61d9b03a8a4e7a407bc2d45e545a960/datasets/util/ade20k/README.md)
```markdown
# ADE20K
<!-- markdown-link-check-disable-next-line -->
ade_20k_calibration_files.txt source is <https://github.com/mlcommons/mobile/blob/master/calibration/ADE20K/ADE20k_calibration.txt>
```

anhappdev:
The mentioned link is here:
https://github.com/mlcommons/mobile_app_open/blob/903bd43c8d2faf5eff0e9104e00a5531f1c11336/CONTRIBUTING.md?plain=1#L14

`403` means forbidden and requires authorization. I remember there are some other links to SDK which also have this error `403` and were excluded from check, but the browser can redirect to login page.
If the resource is not available, we will often get a `404` instead.
So I think it's quite safe to treat `403` as a successful "live link" check.

d-uzlov:
>The mentioned link is here

I mean, in which pull requests or commits did it fail?

anhappdev:
This one: https://github.com/mlcommons/mobile_app_open/pull/292
I rerun the check several times but it always failed.

d-uzlov:
>403 means forbidden and requires authorization

Resources don't tend to suddenly require authorization.
If something is accessible now, I expect this link to remain accessible in the foreseeable future.
If some link is known to require authorization, we can mark it.
If some resource was available freely but started requiring authorization, we may need to audit it and replace the link with something else, or add a note how to get authorized.

anhappdev:
Ok. I will add the exclusion for the mentioned link.

PR #301, opened 2022-05-04
Add new task `IS Mosaic` to Flutter app

### LINKED ISSUE #300: Add new task `IS Mosaic` to Flutter app
There are multiple changes so I created a new issue and PR for this instead of adding it to the big PR #292

anhappdev:
Closes #300

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,7 @@ MLCommons project work is tracked with issue trackers and pull requests. Modify 
 ## Contributing guidelines
 
 If you have improvements to MLPerf, send us your pull requests! For those
+<!-- markdown-link-check-disable-next-line -->
 just getting started, Github has a [howto](https://help.github.com/articles/using-pull-requests/).
 
 ### Pull Request Checklist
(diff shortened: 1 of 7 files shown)


d-uzlov (changes requested):
In the benchmark list in settings the correct names are used but in the results there are 2 seemingly duplicate "Image segmentation tasks" without version specified.

d-uzlov:
I didn't look up the sources for this page, but apparently we use some mapping from benchmark type to benchmark name instead of using name from the tasks file?

Maybe we should merge this and create a separate issue for incorrect naming.
@anhappdev what do you think?

anhappdev:
Yes, currently we show only hard-coded name for 5 different tasks type: IC, IC(offline), OD, IS, LU.

https://github.com/mlcommons/mobile_app_open/blob/149b224fd5b19179bc6acd5418a243e4d7e89372/flutter/lib/benchmark/info.dart#L36-L68

I will make the changes to show the task name from the `task.pbtxt` instead in another PR.

d-uzlov (approved):

PR #305, opened 2022-05-08
Add checksum check for model files

### LINKED ISSUE #295: Handling of case when Git LFS bandwidth limit exceeded
With Git LFS, the existed URL will redirect to the original file so that the download still work as expected.
But when the bandwidth limit exceeded, it redirects to a pointer file. The app still download the file as normal but doesn't know it's a pointer file. We need to check for this case.

Since the pointer file is tiny (about 200 bytes) we can check the file size after download, if it's smaller than a threshold (say 1 KB) the app should show an error.

Example URL: 
https://github.com/mlcommons/mobile_models/raw/main/v2_0/SNPE/deeplabv3_htp.dlc

anhappdev:
Closes #295

### DIFF
--- a/flutter/lib/l10n/app_en.arb
+++ b/flutter/lib/l10n/app_en.arb
@@ -53,7 +53,7 @@
   "errorDialogTitle": "Errors",
   "successDialogTitle": "Success",
   "incorrectDatasetsPath": "Selected datasets directory does not contain files for following benchmarks:",
-  "wrongChecksum": "The checksum of following files are not matched:",
+  "wrongChecksum": "The checksum of the following files are not matched with one defined in benchmark_setting:",
   "ok": "Ok",
   "waitBenchmarks": "Wait for benchmark to finish",
   "cooldownStatus": "Cooldown pause",
--- a/flutter/lib/l10n/app_en.arb
+++ b/flutter/lib/l10n/app_en.arb
@@ -53,7 +53,7 @@
   "errorDialogTitle": "Errors",
   "successDialogTitle": "Success",
   "incorrectDatasetsPath": "Selected datasets directory does not contain files for following benchmarks:",
-  "wrongChecksum": "The checksum of the following files are not matched with one defined in benchmark_setting:",
+  "wrongChecksum": "The following files failed checksum validation:",
   "ok": "Ok",
   "waitBenchmarks": "Wait for benchmark to finish",
   "cooldownStatus": "Cooldown pause",
--- a/flutter/lib/l10n/app_en.arb
+++ b/flutter/lib/l10n/app_en.arb
@@ -53,6 +53,7 @@
   "errorDialogTitle": "Errors",
   "successDialogTitle": "Success",
   "incorrectDatasetsPath": "Selected datasets directory does not contain files for following benchmarks:",
+  "wrongChecksum": "The checksum of following files are not matched:",
   "ok": "Ok",
   "waitBenchmarks": "Wait for benchmark to finish",
   "cooldownStatus": "Cooldown pause",
--- a/flutter/lib/resources/cache_manager.dart
+++ b/flutter/lib/resources/cache_manager.dart
@@ -69,7 +69,7 @@ class CacheManager {
       final archivePath = getArchive(resource);
       if (archivePath == null) continue;
       final archiveFile = File(archivePath);
-      if (archiveFile.existsSync()) await archiveFile.delete();
+      if (await archiveFile.exists()) await archiveFile.delete();
     }
   }
 
--- a/flutter/lib/resources/cache_manager.dart
+++ b/flutter/lib/resources/cache_manager.dart
@@ -20,6 +20,14 @@ class CacheManager {
     return _resourcesMap[url];
   }
 
+  String? getArchive(String url) {
+    if (!isResourceAnArchive(url)) return null;
+    final archiveDirPath = get(url);
+    if (archiveDirPath == null) return null;
+    final archiveFilePath = archiveDirPath + ArchiveCacheHelper.extension;
+    return archiveFilePath;
+  }
+
   Future<void> deleteLoadedResources(List<String> nonRemovableResources,
       [int atLeastDaysOld = 0]) async {
     final directory = Directory(loadedResourcesDir);
@@ -53,6 +61,15 @@ class CacheManager {
         }
       }
       await file.delete(recursive: true);
+    }
+  }
+
+  Future<void> deleteArchives(List<String> resources) async {
+    for (final resource in resources) {
+      final archivePath = getArchive(resource);
+      if (archivePath == null) continue;
+      final archiveFile = File(archivePath);
+      if (archiveFile.existsSync()) await archiveFile.delete();
     }
   }
 
--- a/flutter/lib/resources/resource_manager.dart
+++ b/flutter/lib/resources/resource_manager.dart
@@ -9,6 +9,7 @@ import 'package:yaml/yaml.dart';
 
 import 'package:mlperfbench/device_info.dart';
 import 'cache_manager.dart';
+import 'resource.dart';
 import 'result_manager.dart';
 import 'utils.dart';
 
@@ -21,20 +22,6 @@ class BatchPreset {
     required this.name,
     required this.batchSize,
     required this.shardsCount,
-  });
-}
-
-enum ResourceTypeEnum { model, lib, datasetData, datasetGroundtruth }
-
-class Resource {
-  final String path;
-  final ResourceTypeEnum type;
-  final String? md5Checksum;
-
-  Resource({
-    required this.path,
-    required this.type,
-    this.md5Checksum,
   });
 }
 
@@ -220,7 +207,7 @@ class ResourceManager {
       } else {
         localPath = cacheManager.get(resource.path);
       }
-      if (localPath == null || !File(localPath).existsSync()) continue;
+      if (localPath == null || !(await File(localPath).exists())) continue;
       if (!await isChecksumMatched(localPath, md5Checksum)) {
         checksumFailedResources.add(resource);
       }
--- a/flutter/lib/resources/resource_manager.dart
+++ b/flutter/lib/resources/resource_manager.dart
@@ -8,6 +8,7 @@ import 'package:path_provider/path_provider.dart';
 import 'package:yaml/yaml.dart';
 
 import 'package:mlperfbench/device_info.dart';
+import 'archive_cache_helper.dart';
 import 'cache_manager.dart';
 import 'result_manager.dart';
 import 'utils.dart';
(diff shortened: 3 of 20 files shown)


anhappdev:
The `md5Checksum` field is filled for TFLite backend. I will fill the values for other backends after we reviewed the implementation.

Currently, I use the Flutter package [md5_file_checksum](https://pub.dev/packages/md5_file_checksum) and it supports only iOS and Android. But it should be easy to modify the package to support Windows as well later.

freedomtan:
Looks good to me. nitpick: I prefer to see md5 stored on server side.
For MD5, there are several open source implementations. At least, we can use Ron Rivest's C code in [RFC1321](https://www.ietf.org/rfc/rfc1321.txt)

anhappdev:
>  I prefer to see md5 stored on server side.

It depends on the objective. 
For checking if the downloaded file is a pointer or corrupted, storing the checksum in code would be enough, IMO. 
For checking the file against manipulation, we need to invest more effort to think about a throughout solution.

flutter/lib/l10n/app_en.arb
```diff
@@ -53,6 +53,7 @@
   "errorDialogTitle": "Errors",
   "successDialogTitle": "Success",
   "incorrectDatasetsPath": "Selected datasets directory does not contain files for following benchmarks:",
+  "wrongChecksum": "The checksum of the following files are not matched with one defined in benchmark_setting:",
```
d-uzlov, line 56: I think the following would sound better:
>The following files failed checksum validation

d-uzlov:
I think we should validate model checksums when downloading. It seems very strange to me that you could open the app, download everything, then go somewhere where is no internet and find out that some of the files are invalid. Would be much better if the you would see the error at the time of initial download.

Validating checksums before starting benchmarks accomplishes a bit different task, it ensures that model files were not edited by user. I don't know if we need it but I guess it will not hurt.

d-uzlov (changes requested):
>Currently, I use the Flutter package [md5_file_checksum](https://pub.dev/packages/md5_file_checksum) and it supports only iOS and Android. But it should be easy to modify the package to support Windows as well later.

Currently this PR crashes on Windows because of this.

We should at least disable the validation when running on Windows.

Ideally we would use something else to calculate checksums on Windows.
Maybe we could check current platform to fallback to a less optimized md5 implementation when running on Windows.

anhappdev:
> Currently this PR crashes on Windows because of this.

Maybe we can implement or use an algorithm to calculate MD5 in C/C++ then use `dart:ffi` to call it? What do you think about this? I have no experience with Windows so I don't know if there is any potential issues with Windows using this method.

d-uzlov:
>Maybe we can implement or use an algorithm to calculate MD5 in C/C++ then use dart:ffi to call it?

Should theoretically work fine. We already have native code integration so I don't think there would be any major issues, assuming C code itself doesn't have issues.
Code from that RFC1321 from 1992 that freedom mentioned could have troubles compiling on Windows if there are any linux-specific dependencies but I guess there are enough md5 libraries and examples to pick something platform-independent.

anhappdev:
> Should theoretically work fine.

Ok. I will implement it in a new PR since it may take some time to finish but we need the checksum validation now since we have already switched to Git LFS.

flutter/lib/l10n/app_en.arb
anhappdev: Done

anhappdev:
I found another Flutter package (https://pub.dev/packages/crypto) which has support for multiple platforms. It's  implemented in pure Dart so it's a bit slower than the native code version but still acceptable IMO. We will fix it later if performance is an issue.

d-uzlov:
I would say that `crypto` is _significantly_ slower than the platform native implementation.
In previous version I didn't even notice that anything changed when launching benchmark (I added a timer to the validation functions, and it seems like total time was about 20 ms, or something) but with the last commit of this PR my phone hangs for 2-3 seconds before actually starting the benchmark.

I think we should import both libraries, and only use crypto on Windows.

d-uzlov:
By the way, I think there will be problems with the "validate just before starting" approach.
This validation will only work for single files.
The app also allows you to specify a link to archive. The app will download the archive, unpack it and delete the original file.

Currently none of the backends use archives for models but we discussed that Intel backend uses several files per model, so they will use archives.

We can't specify checksums for each individual file inside an archive, we can validate only the archive as a whole. But we don't keep archives. If we were to keep them, app data folder would grow by 2+ times, and all the archives would just be dead weight. And validating it wouldn't mean anything unless we also deleted everything extracted from archives and unpacked them from scratch each time before starting benchmarks. Which sounds ridiculous to me.

I think the right solution here is to validate files right after downloading, before unpacking. Validating that user didn't change anything is a complex task, separate from the issues with Git LFS that this PR is trying to solve, I don't think it worth trying to solve it here, without a complete package of measures.

d-uzlov:
And if we only validate downloaded files once after the download there will be no issues with worse performance of the hashing algorythm.

anhappdev:
Ok. I didn't think about the case with archive file. So in this case, validate the checksum only one after download is indeed better approach.

flutter/lib/resources/resource_manager.dart
```diff
@@ -71,26 +86,42 @@ class ResourceManager {
...
+    final internetPaths = internetResources.map((e) => e.path).toList();
+    await cacheManager.cache(internetPaths, (double val) {
       _progressString = '${(val * 100).round()}%';
       _onUpdate();
     }, purgeOldCache);
 
+    final checksumFailed = await validateResourcesChecksum(resources);
+    if (checksumFailed.isNotEmpty) {
+      final mismatchedPaths = checksumFailed.map((e) => '\n${e.path}').join();
+      throw 'Checksum validation failed for: $mismatchedPaths';
```
anhappdev, line 119: I moved the checksum validation into the `handleResources()` method. It will throw an exception if checksum validation failed. I am expected it will show the `ExceptionWidget` when the exception is thrown but the app just crashes. The same for the case `forbidden path` exception some lines above.

@d-uzlov Do you have an idea why the app does not show the exception message in the UI, and how can we fix it?

d-uzlov: `resourceManager.handleResources` is an async function without `Future` object.
We don't `await` it, so the exception that is thrown inside of it is received by the Flutter framework itself, and it doesn't know how to handle it.

d-uzlov: I think we should be able to wrap `handleResources` in a try-catch block (maybe create a wrapping function) that would create error dialogues when an exception is thrown from `handleResources` or from any of the functions that it calls.

I think that's actually pretty major issue because there are other things that rely of exceptions. It's bad that the app could crash without an error message when anything goes wrong in the `handleResources` function.

d-uzlov: Currently when `ResourceManager` does anything it updates `BenchmarkState` and calls `_onUpdate()`.
We can probably keep it this way, and add some additional field, like `pendingException`.

d-uzlov: I'm not sure how hard it would be to fix it exactly. Maybe we should create a separate issue for this.

anhappdev: It seems that handle the exception in `async` function will takes time so I created new issue for it https://github.com/mlcommons/mobile_app_open/issues/307.

flutter/lib/resources/cache_manager.dart
```diff
@@ -56,6 +64,15 @@ class CacheManager {
     }
   }
 
+  Future<void> deleteArchives(List<String> resources) async {
+    for (final resource in resources) {
+      final archivePath = getArchive(resource);
+      if (archivePath == null) continue;
+      final archiveFile = File(archivePath);
+      if (archiveFile.existsSync()) await archiveFile.delete();
```
d-uzlov, line 72: Is there a reason to use `existsSync` here instead of `exists`?

flutter/lib/resources/resource_manager.dart
```diff
@@ -23,6 +24,20 @@ class BatchPreset {
   });
 }
 
+enum ResourceTypeEnum { model, lib, datasetData, datasetGroundtruth }
```
d-uzlov, line 27: What is `lib` type for?
As far as I see right now it's not used anywhere.

flutter/lib/resources/resource_manager.dart
```diff
@@ -23,6 +24,20 @@ class BatchPreset {
...
+
+class Resource {
+  final String path;
+  final ResourceTypeEnum type;
```
d-uzlov, line 31: What is the purpose of this field?
I kinda like the idea of having it, it could maybe be useful, but right now we set it but never read it in the program.

flutter/lib/resources/resource_manager.dart
```diff
@@ -167,13 +198,33 @@ class ResourceManager {
...
+    for (final resource in resources) {
+      final md5Checksum = resource.md5Checksum;
+      if (md5Checksum == null || md5Checksum.isEmpty) continue;
+      String? localPath;
+      if (cacheManager.isResourceAnArchive(resource.path)) {
+        localPath = cacheManager.getArchive(resource.path);
+      } else {
+        localPath = cacheManager.get(resource.path);
+      }
+      if (localPath == null || !File(localPath).existsSync()) continue;
```
d-uzlov, line 223: Let's use the async `exists` if possible.

flutter/lib/resources/utils.dart
```diff
@@ -1,11 +1,13 @@
+import 'package:mlperfbench/resources/resource_manager.dart';
```
d-uzlov, line 1: I think it would be better to move `Resource` definition into a separate file.
It feels wrong to me to import such unrelated file just for one small data class.

flutter/lib/resources/cache_manager.dart
anhappdev: We can use `await file.exists()` here but I don't see any difference.

d-uzlov: `existsSync` blocks the thread.
Since we are running almost everything in the UI thread blocking it makes the UI unresponsive.
Well, most of the time it will probably be fine, but if the storage does not respond for any reason delays could get noticeable.

d-uzlov: It's not a major issue by any means but I don't see any reason to take additional risks.

flutter/lib/resources/resource_manager.dart
anhappdev: In the first implementation I use it to check if the resource is a model file and run checksum validation only against model files. But now I just run checksum validation against any files which have a valid `md5checksum`.
We can remove `ResourceTypeEnum type` but I also think it will be helpful to know the type of the resource.

flutter/lib/resources/resource_manager.dart
anhappdev: ok

flutter/lib/resources/utils.dart
anhappdev: ok

flutter/lib/resources/resource_manager.dart
anhappdev: I have seen in the code somewhere that the app can also have a path to a lib file, so I add it, but it is not used yet, we can remove it and add it later if needed.

flutter/lib/resources/cache_manager.dart
anhappdev: Ok.

d-uzlov (approved):

PR #309, opened 2022-05-17
Make `single_stream_expected_latency_ns` configurable in benchmark_setting for Flutter app

### LINKED ISSUE #291: Make expected_latency_ns a backend configurable parameter
Different backends require to set this differently to spend at least 60sec in the single stream benchmark. It should be responsibility of backend to set this correct to get a VALID submission. The default parameter of 1M had to be lowered to 900k or 800k in different v2 submissions. Is better if we set the default parameter as 1M and backend can overwrite if needed 
[mlperf_settings.single_stream_expected_latency_ns](https://github.com/mlcommons/mobile_app_open/blob/submission-v2.0/android/cpp/mlperf_driver.cc#L105)

anhappdev:
Closes #291


flutter/cpp/proto/backend_setting.proto
```diff
@@ -47,6 +47,8 @@ message BenchmarkSetting {
   required string configuration = 6;
   // batch_size is passed to the backend
   optional int32 batch_size = 7;
+  // single_stream_expected_latency_ns is passed to the backend, default to 1M
+  optional int32 single_stream_expected_latency_ns = 8 [default = 1000000];
```
anhappdev, line 51: In this implementation the `single_stream_expected_latency_ns` will be defaulted to 1M.

mobile_back_tflite/cpp/backend_tflite/tflite_settings_android.h
```diff
@@ -96,6 +96,7 @@ benchmark_setting {
   accelerator_desc: "NNAPI"
   configuration: "TFLite"
   src: "https://github.com/mlcommons/mobile_models/raw/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_uint8.tflite"
+  single_stream_expected_latency_ns: 900000
```
anhappdev, line 99: The vendor can change the default value of 1M by explicitly set it in the `benchmark_setting`

mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/eden_sdk/include/samsung_settings.h
```diff
@@ -96,6 +96,7 @@ benchmark_setting {
   accelerator_desc: "npu"
   configuration: "Samsung Exynos"
   src: "https://github.com/mlcommons/mobile_models/raw/main/v1_0/Samsung/is.nnc"
+  single_stream_expected_latency_ns: 900000
```
anhappdev, line 99: The vendor can change the default value of 1M by explicitly set it in the benchmark_setting

d-uzlov (approved):

freedomtan:
Let's wait samsung backend changes.

freedomtan:
@mohitmundhragithub and other Qualcomm colleagues, please help review this and change q backend for this latency parameter accordingly.

mobile_back_samsung/cpp/eden_mlperf_c_backend/jni/eden_sdk/include/samsung_settings.h
Mostelk: Group policy discussion is we need to enforce that this parameter does not exceed 1M, to avoid manipulation for slower tasks, please modify the PR to take min(parameter, 1M) @anhappdev

anhappdev: Should the app throw an exception if the latency > 1M (which I prefer) or just silently take min(parameter, 1M)?

Mostelk: I think just silently take min(parameter, 1M) to avoid disruption, and in settings file we can put a comment to reflect that

anhappdev: If the app just silently takes min(parameter, 1M), other / new engineers who not know about this logic will wonder why latency > 1M has no effect. Each backend has a separated setting file, so there will be many comments to add and (if they are noted is another question).
Throw an exception would be a more correct way to handle this, IMO. If someone wants to change the default latency, they will run a test to see how it works and knows exactly what to fix if the latency > 1M.

Mostelk: Yes, throwing an exception sounds more correct.

anhappdev: Done

d-uzlov:
> new log location

On Android Flutter app uses system-provided cache directory.
It prints it into logcat as `output_dir=<path>`. On my phone path is `/data/user/0/org.mlcommons.android.mlperfbench/cache`.
It seems like this path is not accessible on non-rooted devices so we may think about moving logs to somewhere else.

However, about a month ago we merged a PR that reads logs and marks invalid results in UI by red color. If results in UI use black/white color for text then loadgen marked them as valid.

d-uzlov:
Also loadgen short summary should be printer into logcat.

mohitmundhragithub:
Thanks a lot @d-uzlov  for the steps. I was able to verify that this change is working fine for Qualcomm.

mohitmundhragithub (dismissed):

mohitmundhragithub (dismissed):

anhappdev:
@d-uzlov Please review and approve as I needed to merge master to fix CI checks.

anhappdev:
I forgot to delete it.

Issue #306, opened 2022-05-09
Easy way to detect invalid results


mohitmundhragithub:
We should show some indication on the UI that the results are valid or not. During the last submission, we some invalid results caused unnecessary work for SS and QC at the last moment.
Some indication on the UI (say colour change of the case which is invalid or some error message), will definitely help in avoiding such unnecessary work.

freedomtan:
We can check result json file to if a benchmark result is valid or not. Dunno if loadgen returns valid  or not in its API.

@d-uzlov please help this after windows related ones.

PR #312, opened 2022-05-19
Mark invalid results

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/306

### DIFF
--- a/flutter/lib/backend/bridge/isolate.dart
+++ b/flutter/lib/backend/bridge/isolate.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 import 'dart:isolate';
 
 import 'package:mlperfbench/backend/bridge/ffi_run.dart';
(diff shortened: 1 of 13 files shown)


d-uzlov:
This is how the result page looks with this change:

Results which loadgen marked as `valid` in logs are black, `invalid` are red.
If a benchmark did not run it is also marked as red.

flutter/lib/backend/bridge/isolate.dart
```diff
@@ -48,7 +49,27 @@ class BridgeIsolate {
     sendPort.send(port.sendPort);
     await for (var settings in port.cast<RunSettings>()) {
       var result = runBenchmark(settings);
+      result.validity = await isResultValid(settings);
       sendPort.send(result);
     }
   }
+
+  static Future<bool> isResultValid(RunSettings settings) async {
```
anhappdev, line 57: Can you add a comment or relevant link to explain why we use this logic to check for validity? In case the logic is changed by external party we follow and can update it.

anhappdev (changes requested):

flutter/lib/backend/bridge/isolate.dart
d-uzlov: Done

anhappdev (approved):

PR #314, opened 2022-05-19
Refactor pixel backend by following the changes in tflite backend

kcwu:
- Move interpreter resize to initialization stage
 - Add check for batch/shard compatibility. Format, optimize variables

With this change, pixel backend won't create unused interpreters.


freedomtan (approved):
LGTM

Issue #319, opened 2022-05-25
Save benchmark result history


d-uzlov:
Parent issue: https://github.com/mlcommons/mobile_app_open/issues/253

Currently the app only keeps the last result.
When you rerun benchmarks then all previous results are erased. Also, if you cancel the benchmark in the middle of the run results are also erased.

We want to modify the UI to show previous results in the future so we should change this behavior.
We can show only the last result until we properly modify the UI but the backend should keep all results.

We can keep current behavior of "remove results from the UI if the run was cancelled".

PR #323, opened 2022-05-26
Save benchmark result history

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/319

### DIFF
--- a/flutter/integration_test/first_test.dart
+++ b/flutter/integration_test/first_test.dart
@@ -1,16 +1,14 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
-import 'package:mlperfbench_common/data/extended_result.dart';
 import 'package:shared_preferences/shared_preferences.dart';
-import 'dart:io';
-import 'dart:convert';
 
 import 'package:mlperfbench/main.dart' as app;
 import 'package:mlperfbench/ui/main_screen.dart';
 import 'package:mlperfbench/ui/result_screen.dart';
 import 'package:mlperfbench/resources/resource_manager.dart'
     as resource_manager;
+import 'package:mlperfbench/resources/result_manager.dart' as result_manager;
 
 void main() {
   final splashPauseSeconds = 4;
@@ -67,16 +65,12 @@ void main() {
 
       final applicationDirectory =
           await resource_manager.ResourceManager.getApplicationDirectory();
-      final jsonResultPath = '$applicationDirectory/result.json';
-      final file = File(jsonResultPath);
 
-      expect(await file.exists(), true,
-          reason:
-              'Result.json does not exist: file $applicationDirectory/result.json is not found');
+      final rm = result_manager.ResultManager(applicationDirectory);
+      await rm.init();
 
-      final jsonResultContent = await file.readAsString();
-      final extendedResults = ExtendedResult.fromJson(
-          jsonDecode(jsonResultContent) as Map<String, dynamic>);
+      final extendedResults = rm.getLastResult();
+      ;
       final length = extendedResults.results.list.length;
 
       expect(length, expectedResultCount,
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -158,6 +158,7 @@ class BenchmarkState extends ChangeNotifier {
       try {
         lastResult = ExtendedResult.fromJson(
             jsonDecode(_store.previousExtendedResult) as Map<String, dynamic>);
+        // TODO: restore from resultManager.getLastResult ?
         resourceManager.resultManager
             .restoreResults(lastResult!.results.list, benchmarks);
         return true;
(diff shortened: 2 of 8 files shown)


anhappdev (changes requested):
All the results are stored in one big JSON file instead of one JSON file per result, it's convenience but:
1.  if the file is corrupted, all data is gone
2.  if the file is big enough, parsing and writing will be a problem
3.  if the structure of the JSON file is changed in the feature, parsing will be a problem
4.  if the user needs to delete one result for any reason, delete just one file will be easier than modify one big file

I would store each result in a separated JSON file with a timestamp in the file name.

mkankana:
@anhappdev @d-uzlov  if we decide to save in separate files , Can I know 
1. How many of the history result will be saved ? will it be all the runs ?
2. Will there be an option to delete all or one of the saved history result files ?

d-uzlov:
> 1. if the file is corrupted, all data is gone

I don't see why a file could become corrupted, unless user incorrectly edited it, which I don't think users should do.

> 3\. if the structure of the JSON file is changed in the feature, parsing will be a problem

I don't think this is related to the way we store results on disk. Changes in structure will be a challenge with any storage method.

> 4\. if the user needs to delete one result for any reason, delete just one file will be easier than modify one big file

From the app side there is no difference.

> 2\. The history files should be easily deleted by using any File Manager app.

I don't think users should be able to edit results manually.

> 3\. Each result is just a JSON file, so the user can share it by using any external app like Mail, Message, etc. If we need a UI to share it directly from the app, this should be discussed and prioritized.

I don't think it's any different from the "Share Results" button that we already have in the app.
We could discuss details for "Share" button in result history after we at least have some UI for result history, which is not included in this PR.

d-uzlov:
I agree that using JSON for storing results is not the best choice.
But I think that creating a bunch of JSON files on disk and asking users to manipulate them is much worse.

The main issues that I see with using a JSON file are:

* Users could edit it in an unwanted way. It's possible to alter results or simply corrupt file
  * Well, it's not like only JSON files have these issues but any binary file is better in this regard
* It's not supposed to work with big files
  * But big is pretty big in our case. File with 1k results is about 5 MB in size and takes about 50 milliseconds to parse or serialize
  * I don't think real users are likely to have any issues

And if users are not supposed to edit results on disk manually there are no issues from usability standpoint. With any serialization method.

I think using JSON is fine as a temporary solution for development versions of the app.

---

I think the proper solution here is to use a proper database instead of a JSON file. Like [hive](https://pub.dev/packages/hive) or [objectbox](https://pub.dev/packages/objectbox).

But I think it would be better to do this in a separate PR. Choosing and integrating a proper database could take some time and could require discussing something if there are any issues.
This PR contains important changes that allow to save and restore results, and to access them in the app. The actual serialization is a very minor part of this PR.

anhappdev:
1. File corruption is not common but it happens for numerous reasons. By storing many small files we reduce the risk of losing all data in one big file.
3. The user should not edit the JSON file (else it's data manipulation) but may want to easily delete a certain result / file.
2. A change in the data structure may require a migration in a real DB, but if we have separated JSON files, it's easier to handle IMO.
4. The idea (as you have also pointed out) is to **quickly** implement the feature and reuse what's already there (File Manager, Mail app, etc.). We can implement a full-featured CRUD app / UI for this but it's take many more efforts and need to be prioritized.

d-uzlov:
> 1. File corruption is not common but it happens for numerous reasons. By storing many small files we reduce the risk of losing all data in one big file.

App executable can become corrupt.
App native libraries can become corrupt.
Model files can become corrupt, and the app will crash.

I don't see how a system where files can spontaneously become corrupt could be usable so I don't think it should be our concern.

> 2\. The user should not edit the JSON file (else it's data manipulation) but may want to easily delete a certain result / file.

I don't see why user would need or want to delete files manually if there will be "Delete" button in the UI.

> 3\. A change in the data structure may require a migration in a real DB, but if we have separated JSON files, it's easier to handle IMO.

* If we use a real DB and we don't have a migration:
  * If we delete old results, we will lose all data.
  * If we don't delete old results, we will have dead weight (until we have a migration)
* If we use several JSON files and we don't have a migration:
  * If we delete old results, we will lose all data.
  * If we don't delete old results, we will have dead weight (until we have a migration)

We don't need a **_migration_** for a nosql DB to store new changed data. We only need it to access results in the old format.

The only difference is that dead weight contained in JSON files could be recovered by user by manually finding them.
But I don't think users should ever touch files that the app creates for its internal usage.

> We can implement a full-featured CRUD app / UI for this but it's take many more efforts and need to be prioritized.

We don't need a full-blown CRUD for this.
We only ever need to add/remove whole results.
We must never edit anything inside a result.
Showing results to user doesn't mean showing all the fields from result. UI does not depend on data structure at all.
Migration is not really "updating", it's more of "read everything, delete database, write from scratch". It's also automatic, no need for CRUD here.

And we already have a task to create UI for result history. And it's prioritized.
But it will be UI for history, not CRUD. Because we don't need CRUD.

anhappdev:
User-generated files are easier to be corrupted for example if the app crashes while it's writing the file to disk.
I think the user has intention to use the JSON files outside the app. If not, it should be stored in a location where the user cannot access them, no matter if it's a single or multiple files.
I don't know yet how the UI will be implemented, but a File Manager will be certain more featured to use (batch delete, backup, share, ...) and it's already there.

PR #325, opened 2022-05-26
Add or modify IMAGEPAIRS related files

JaehaKim97:
- Add new dataset type "IMAGEPAIRS".

- We've tested these codes with 128 x 128 input images with SR scale factor 2 and verified that the result of PSNR is the same as conventional PSNR measuring code written by python.

- However, the current code does not support png type image which is commonly used in the SR field. We will continuously update the code.


flutter/pubspec.lock
```diff
@@ -28,7 +28,7 @@ packages:
       name: archive
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "3.1.6"
+    version: "3.1.2"
```
d-uzlov, line 31: Please, revert changes in `pubspec.lock`.

d-uzlov (changes requested):

flutter/pubspec.lock
JaehaKim97: We have reverted changes in `pubspec.lock`.

JaehaKim97:
Hi @d-uzlov,

I have tried to match the format of `*.cc` and `*.h` files via `clang-format`, but it continuously shows a mismatch.

It seems that 'line ending characters' occur errors, but we can not figure out what exactly causes the problem.

Could you please inform me what the problem is?

d-uzlov:
@JaehaKim97 you probably use different settings for clang-format.
You can run `make format/clang` to fix it.

d-uzlov:
Issue with line endings can be fixed using `make format/line-endings`.
But it seems right now there are no issues with line endings.

d-uzlov (approved):

PR #341, opened 2022-06-03
Use native library in Flutter Android

mohitmundhragithub:
1) With 31 android version, with built with qti
backend, the app crashes, because it's not able
to load the libs from /vendor/lib64 or /system/vendor/lib64
paths. These folders contain dependent libraries.
So, reducing the target SDK version.
2) A change in docker file for file permission.
3) Updated the settings

### DIFF
--- a/flutter/android/app/build.gradle
+++ b/flutter/android/app/build.gradle
@@ -44,7 +44,7 @@ android {
     defaultConfig {
         applicationId "org.mlcommons.android.mlperfbench"
         minSdkVersion 16
-        targetSdkVersion 30
+        targetSdkVersion 31
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
--- a/flutter/android/app/build.gradle
+++ b/flutter/android/app/build.gradle
@@ -44,7 +44,7 @@ android {
     defaultConfig {
         applicationId "org.mlcommons.android.mlperfbench"
         minSdkVersion 16
-        targetSdkVersion 31
+        targetSdkVersion 30
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
--- a/mobile_back_qti/make/qti_backend.mk
+++ b/mobile_back_qti/make/qti_backend.mk
@@ -34,6 +34,7 @@ ifeq (${WITH_QTI},1)
     ${local_snpe_sdk_root}/lib/dsp/libsnpe_dsp_v68_domains_v3_skel.so \
     ${local_snpe_sdk_root}/lib/dsp/libSnpeHtpV69Skel.so \
     ${local_snpe_sdk_root}/lib/dsp/libSnpeHtpV68Skel.so
-  backend_qti_android_target=//mobile_back_qti/cpp/backend_qti:libqtibackend.so
+  backend_qti_android_target=//mobile_back_qti/cpp/backend_qti:libqtibackend.so \
+                             //flutter/android/commonlibs:commonlibs
   backend_qti_filename=libqtibackend
 endif
--- a/mobile_back_qti/make/qti_backend.mk
+++ b/mobile_back_qti/make/qti_backend.mk
@@ -22,7 +22,7 @@ ifeq (${WITH_QTI},1)
   local_snpe_sdk_root=$(shell echo mobile_back_qti/snpe-* | awk '{print $$NF}')
   $(info detected SNPE SDK: ${local_snpe_sdk_root})
   backend_qti_android_files=${BAZEL_LINKS_PREFIX}bin/mobile_back_qti/cpp/backend_qti/libqtibackend.so \
-    ${local_snpe_sdk_root}/lib/aarch64-android-clang8.0/libc++_shared.so \
+    ${BAZEL_LINKS_PREFIX}bin/flutter/android/commonlibs/lib_arm64/libc++_shared.so \
     ${local_snpe_sdk_root}/lib/aarch64-android-clang8.0/libSNPE.so \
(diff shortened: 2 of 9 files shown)


mobile_back_qti/make/qti_backend.mk
```diff
@@ -22,7 +22,7 @@ ifeq (${WITH_QTI},1)
   local_snpe_sdk_root=$(shell echo mobile_back_qti/snpe-* | awk '{print $$NF}')
   $(info detected SNPE SDK: ${local_snpe_sdk_root})
   backend_qti_android_files=${BAZEL_LINKS_PREFIX}bin/mobile_back_qti/cpp/backend_qti/libqtibackend.so \
-    ${BAZEL_LINKS_PREFIX}bin/flutter/android/commonlibs/lib_arm64/libc++_shared.so \
+    ${local_snpe_sdk_root}/lib/aarch64-android-clang8.0/libc++_shared.so \
```
d-uzlov, line 25: Why do we need this change?

This change is incompatible with Samsung backend. 

Samsung backend also requires `libc++_shared.so` file.
If the file provided by SNPE SDK is different from `flutter/android/commonlibs/lib_arm64/libc++_shared.so` it may be incompatible with Samsung backend by itself.
And even if it's compatible there will be troubles building the app because currently we copy all files listed by all backends, and the copy command will complain about 2 files with identical names.

d-uzlov (changes requested):

anhappdev:
With target SDK v30 we will not be able to publish the app. See
https://developer.android.com/google/play/requirements/target-sdk

flutter/android/app/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
     defaultConfig {
         applicationId "org.mlcommons.android.mlperfbench"
         minSdkVersion 16
-        targetSdkVersion 31
+        targetSdkVersion 30
```
anhappdev, line 47: With target SDK v30 we will not be able to publish the app. See
https://developer.android.com/google/play/requirements/target-sdk

anhappdev (changes requested):

flutter/android/app/build.gradle
mohitmundhragithub: but seems like targetSdkVersion 31 is somehow creating issues with flutter. The app is not able to read the files from /system/lib64 folder. So, the system dependent libraries like libcdsprpc.so are not found when loading libqtibackend.so.

My suggestion is to have this changed to 30 and find a solution for the library load issue separately.

mobile_back_qti/make/qti_backend.mk
mohitmundhragithub: somehow flutter/android/commonlibs path was not working with docker flutter steps. 
I will try to see how to map bazel-bin from inside docker.

mohitmundhragithub: found that flutter/commonlibs was not built when building only qti backend. 
So, moved bulding of flutter commonlibs to tflitebackend, which is enabled by default.

flutter/android/app/build.gradle
d-uzlov: I found the following info in Android documentation:
>Applications that target SDK version 31 (Android 12) or higher must explicitly specify their native shared library dependencies by using the <uses-native-library> tag in the app manifest. If any part of the requested library doesn't exist on the device, the app isn't installed. When the apps are installed, they're provided with only the native shared libraries that they've requested. This means that apps can't access native shared libraries that don't appear in the app manifest.

https://source.android.com/devices/tech/config/namespaces_libraries

It seems to me that if we want to create a unified app for all backends we will have to include all vendor-specific libraries into the app, or else the app won't even install anywhere.

mobile_back_qti/make/qti_backend.mk
d-uzlov: Could you move commonlibs bazel target to both `backend_qti_android_target` and `backend_samsung_android_target` configs, please?
Bazel doesn't mind duplicated targets.
And it's technically possible to build the app without tflite backend.

flutter/android/app/build.gradle
anhappdev: Downgrade to v30 is not really a choice if we want to publish the app. We can try what @d-uzlov suggest.

freedomtan:
It seems to me discussion above is mixing Android SDK and NDK versions. The `libc++_shared.so` is in NDK not SDK.

mohitmundhragithub:
> It seems to me discussion above is mixing Android SDK and NDK versions. The `libc++_shared.so` is in NDK not SDK.

There are 2 separate changes. 1) related to libc++_shared.so. 2) related to reducing the sdk version to 30, from current 31.

flutter/android/app/build.gradle
mohitmundhragithub: looking into it if doing that works.

mobile_back_qti/make/qti_backend.mk
mohitmundhragithub: done.

flutter/android/app/build.gradle
d-uzlov: I think we can merge this for now and update the SDK target version later, after we find the solution.

d-uzlov (dismissed):

flutter/android/app/build.gradle
anhappdev: Please discuss it in the mobile engineering meeting how we want to handle it.

d-uzlov: It seems there are no issues with specifying list of vendor-specific native libraries after all.
There is a flag to mark dependency as optional, and with the last commit from @AasheeshT I'm able to install the app on my phone which doesn't have specified Qualcomm native lib.

anhappdev: Also worked with my test device.

anhappdev (approved):

PR #345, opened 2022-06-06
update mediatek backend settings

freedomtan:
use neuron delegate for mosaic
shards_num =  2 -> 4


freedomtan:
this one already merged in #366

Issue #294, opened 2022-04-19
Add support for displaying several accuracy values per benchmark


d-uzlov:
We may need to use 2 different accuracy metrics for some of the future benchmarks, so we need to modify UI to support it.

d-uzlov:
Here are mockups of some possible easy changes:
<details>
<summary>Top results: option  1</summary>
<img width=300 src="https://user-images.githubusercontent.com/36223296/163918794-cc635337-e5a4-4569-973e-66b82cb0a919.png">
</details>

<details>
<summary>Top results: option 2</summary>
<img width=300 src="https://user-images.githubusercontent.com/36223296/163918976-a129fd6d-d4a2-4062-a235-a286bd4d75fe.png">
</details>

<details>
<summary>Bottom results</summary>
<img width=300 src="https://user-images.githubusercontent.com/36223296/163919023-4dff658b-9417-4128-b6fd-c1ccac1f3dd3.png">
</details>

freedomtan:
let's go for option 1 first.

PR #349, opened 2022-06-08
Refactor Flutter native bridge code

d-uzlov:
Part of https://github.com/mlcommons/mobile_app_open/issues/294

This is to make it a bit easier to read and edit bridge functions.
I extracted this to a separate PR to make it easier to review the main PR for the task.

### DIFF
--- a/flutter/cpp/flutter/ios_linker_workaround.cc
+++ b/flutter/cpp/flutter/ios_linker_workaround.cc
@@ -1,0 +1,38 @@
+#include "./ios_linker_workaround.h"
+
+#include <stdint.h>
+
+#include "./dart_backend_match.h"
+#include "./dart_mlperf_config.h"
+#include "./dart_run_benchmark.h"
+#include "flutter/cpp/c/backend_c.h"
+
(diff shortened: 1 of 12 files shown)


flutter/cpp/flutter/ios_linker_workaround.cc
```diff
@@ -0,0 +1,38 @@
...
+#include "./dart_backend_match.h"
+#include "./dart_mlperf_config.h"
+#include "./dart_run_benchmark.h"
+#include "flutter/cpp/c/backend_c.h"
+
+// On iOS we link backend statically and don't export any functions
+// Linker sees that we don't use some of the functions and removes them
+// (when building for release of profile modes)
+// But we still want to access these functions dynamically,
+// so we need fake_calls() function to prevent linker from removing them.
```
anhappdev, line 14: `fake_calls()` is renamed, maybe rewording to:
>so we need to call this function to prevent linker from removing them.

anhappdev (changes requested):

anhappdev (approved):

PR #355, opened 2022-06-08
Fix few issues with accuracy transfer

d-uzlov:
Part of https://github.com/mlcommons/mobile_app_open/issues/294

Fixes for few things I found when adding support for showing second accuracy value:
* App could crash on restoring result history
* Loadgen doesn't give result validity for accuracy runs
* Accuracy is unnecessarily converted between number and string


flutter_common/lib/data/results/benchmark_result.dart
```diff
@@ -3,6 +3,30 @@ import 'backend_settings.dart';
 import 'dataset_info.dart';
 import 'loadgen_scenario.dart';
 
+class AccuracyData {
```
anhappdev, line 6: Just `Accuracy` might be enough.

flutter_common/lib/data/generation_helpers/write_json_example.main.dart
```diff
@@ -28,7 +28,10 @@ const fileNameEnv = 'jsonFileName';
 Future<void> main() async {
   final runResult = BenchmarkRunResult(
     throughput: 123.45,
-    accuracy: 123.45,
+    accuracy: AccuracyData(
+      value: 123.45,
```
anhappdev, line 32: Should the value not be normalized?

flutter/lib/ui/result_screen.dart
```diff
@@ -75,11 +75,11 @@ class _ResultScreenState extends State<ResultScreen>
...
         final throughput = benchmarkResult?.throughput;
         textResult = throughput?.toStringAsFixed(2) ?? 'N/A';
-        numericResult = (throughput ?? 0) / benchmark.info.maxThroughput;
+        numericResult = (throughput ?? 0.0) / benchmark.info.maxThroughput;
       } else if (_screenMode == _ScreenMode.accuracy) {
         benchmarkResult = benchmark.accuracyModeResult;
-        textResult = benchmarkResult?.accuracy ?? 'N/A';
-        numericResult = benchmarkResult?.numericAccuracy ?? 0.0;
+        textResult = benchmarkResult?.accuracy?.string ?? 'N/A';
+        numericResult = (benchmarkResult?.accuracy?.value ?? 0.0) / 100.0;
```
anhappdev, line 82: If we normalize the accuracy between `0.0` and `1.0` then there is no need to have
`virtual float ComputeNormalizedAccuracy() { return ComputeAccuracy() * 100; }`
and then `value / 100.0` here

flutter_common/lib/data/results/benchmark_result.dart
```diff
@@ -3,6 +3,30 @@ import 'backend_settings.dart';
...
+  static const String _tagValue = 'value';
+  static const String _tagString = 'string';
+
+  final double value;
+  final String string;
+
+  AccuracyData({
+    required this.value,
```
anhappdev, line 14: If the accuracy value is normalized, should we check if the value is in the normalized range and raise an exception if not?

anhappdev (changes requested):

flutter_common/lib/data/generation_helpers/write_json_example.main.dart
d-uzlov: It doesn't matter here, we only need this to mark that the value is numeric

flutter_common/lib/data/results/benchmark_result.dart
d-uzlov: I don't think we should check such things in the data class.
Requirement for normalization is a weak requirement. We need it to make value loop pretty in the UI but it doesn't really matter otherwise. From data class standpoint there is no difference.

Also I don't see how we would handle this exception. I think crashing the app because there is an error related to accuracy calculation is a very bad idea.

Maybe we could check that result doesn't exceed the upper bound, and if it does then mark result invalid (red). I'll add this check.

flutter/documentation/result-spec.md
```diff
@@ -61,10 +61,13 @@ If you enable Submission mode, both `performance_run` and `accuracy_run` values
   * `throughput`: floating point number  
     Throughput value for this run of the benchmark.
     May be null for an accuracy run.
-  * `accuracy`: floating point number  
-    Accuracy value for this run of the benchmark.
-    Value must be normalized between `0` and `100`.
+  * `accuracy`: map
     May be null for a performance run if groundtruth file is not provided.
+    * `value`: floating point number
```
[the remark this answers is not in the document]
d-uzlov: I don't think that being normalized is a defining property of this field. It's not like we have `normalizedValue` and `nonNormalizedValue` fields, there is only one numeric field.
`string` value for accuracy is not a "value that may not be normalized", it's a human-readable string value. Normalization isn't even applicable to this field because it's non-numeric.

> I also think it's more common to normalize the value between `0.0` and `1.0`.

Ok, seems like a good idea to me.

flutter_common/lib/data/results/benchmark_result.dart
d-uzlov: > Maybe we could check that result doesn't exceed the upper bound, and if it does then mark result invalid (red). I'll add this check.

Done

flutter/lib/ui/result_screen.dart
d-uzlov: Done

flutter_common/lib/data/results/benchmark_result.dart
d-uzlov: Done

flutter/documentation/result-spec.md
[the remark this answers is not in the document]
d-uzlov: > I also think it's more common to normalize the value between `0.0` and `1.0`.

Done

flutter_common/lib/data/generation_helpers/write_json_example.main.dart
anhappdev: It would be nice if the example value match the description of the key.

flutter/documentation/result-spec.md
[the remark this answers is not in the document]
anhappdev: > Accuracy value for this run of the benchmark.
       Value must be normalized between `0.0` and `1.0`.

The idea is to know that we use normalized value without reading the specs.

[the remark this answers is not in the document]
d-uzlov: > I would explicitly name this `normalizedValue`

Done.

Though, I still don't see why this would be useful.

flutter/lib/backend/bridge/ffi_run.dart
```diff
@@ -115,7 +117,8 @@ RunResult runBenchmark(RunSettings rs) {
 
   var res = RunResult(
     throughput: 1000.0 / runOut.ref.latency,
-    accuracy: runOut.ref.accuracy.toDartString(),
+    accuracyNum: runOut.ref.accuracyNum,
```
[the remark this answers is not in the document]
d-uzlov: Done

flutter_common/lib/data/generation_helpers/write_json_example.main.dart
d-uzlov: Done

flutter_common/lib/data/generation_helpers/write_json_example.main.dart
```diff
@@ -29,8 +29,8 @@ Future<void> main() async {
   final runResult = BenchmarkRunResult(
     throughput: 123.45,
     accuracy: Accuracy(
-      value: 123.45,
-      string: '123.45%',
+      normalized: 0.123,
+      formatted: '123.45%',
```
anhappdev, line 33: `0.123` would be `12.3%`.

d-uzlov: Done.

I don't see the point, though. These values never-ever go our of this file. The string may as well be empty and all numeric values may as well be zero or anything else. This is only for automatic json schema generation, it was never meant to be read by anyone.

anhappdev (approved):

PR #364, opened 2022-06-14
Add support for second accuracy in UI

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/294

Currently no dataset supports 2 accuracy values so second accuracy is disabled.
When we have a dataset that provides a second accuracy value enabling it in UI will be very simple.


d-uzlov:
This PR adds the ability to show second accuracy value but doesn't actually change anything in the UI because currently we don't have any datasets that produce 2 accuracy values.

For these screenshots I artificially added second accuracy to each dataset as a copy of the first.

<details>
<summary>Current looks</summary>


</details>

anhappdev (approved):

PR #366, opened 2022-06-15
Make neuron backend work

freedomtan:
1. use Neuron Delegate in `tensorflow-lite-neuron.aar` for now (will use source code neuron delegate repo when it's ready later)
2. for Android API level 31 and later, we need `uses-native-library` when using underlying vendor libs
3. fix some neuron delegate settings
4. ~~since we build and use protobuf, we can build `protoc` from protobuf instead of depending on system package~~

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -55,12 +55,13 @@ android_workspace()
 #    name = "androidndk",
 #)
 
-http_archive(
-    name = "neuron_delegate",
-    sha256 = "2e4600c99c9b4ea7a129108cd688419eeef9b2aeabf05df6f385258e19ca96c4",
-    strip_prefix = "tflite-neuron-delegate-2.6.0",
-    urls = ["https://github.com/MediaTek-NeuroPilot/tflite-neuron-delegate/archive/v2.6.0.tar.gz"],
-)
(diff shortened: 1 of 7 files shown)


freedomtan:
@d-uzlov could we add `numpy` to the environment of `Form & lint / dart` or I have to revert the building protoc patch?

d-uzlov:
@freedomtan, is there a benefit in building protobuf locally?
Well, except for not having to install it to environment.

Adding numpy wouldn't be hard but I think it would be a bit inconvenient if linter checks required 5-10 minutes instead of current 2 minutes (because github CI doesn't have bazel cache).

freedomtan:
> @freedomtan, is there a benefit in building protobuf locally? Well, except for not having to install it to environment.
> 
> Adding numpy wouldn't be hard but I think it would be a bit inconvenient if linter checks required 5-10 minutes instead of current 2 minutes (because github CI doesn't have bazel cache).

Yes, supposedly, protobuf is version agnostic most of time. However, it seems sometime a system-wide protoc may cause some problems. So some packages come with specific "tested" protobuf version.

I'll revert this one so that we won't need longer lint time.

d-uzlov (approved):

PR #367, opened 2022-06-15
Set up Xcode Cloud

### LINKED ISSUE #121: Investigate Xcode Cloud build
Can we use xcloud build for iOS CI?

anhappdev:
Closes #121


anhappdev:
I will update the workflow to be triggered by new commit in the `master` branch later after we merge this PR.

flutter/ios/ci_scripts/ci_post_clone.sh
```diff
@@ -0,0 +1,45 @@
...
+mkdir -p "$BUILD_HOME"
+test ! -d "$FLUTTER_HOME" && git clone --branch 2.10.5 --depth 1 https://github.com/flutter/flutter.git "$FLUTTER_HOME"
+export PATH="$PATH:$FLUTTER_HOME/bin:$HOME/.pub-cache/bin"
+dart pub global activate protoc_plugin
+
+echo "$LOG_PREFIX Build app"
+cd "$ROOT_DIR" && make
+cd "$ROOT_DIR"/flutter && flutter precache --ios
+cd "$ROOT_DIR"/flutter/ios && pod install
+cd "$ROOT_DIR"/flutter && flutter build ios --config-only integration_test/first_test.dart
```
d-uzlov, line 43: The build uploaded to testflight also launches the test.
We can probably differentiate release and test builds if we create 2 separate xcode cloud workflows and set different environment variables.

As an alternative we can just disable uploading to testflight and do it manually.
But I would prefer to have automatic uploads, they are nice to have.

flutter/ios/ci_scripts/ci_post_clone.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "$LOG_PREFIX Install Flutter"
+export BUILD_HOME="$HOME"/build
+export FLUTTER_HOME="$BUILD_HOME"/flutter
+mkdir -p "$BUILD_HOME"
+test ! -d "$FLUTTER_HOME" && git clone --branch 2.10.5 --depth 1 https://github.com/flutter/flutter.git "$FLUTTER_HOME"
+export PATH="$PATH:$FLUTTER_HOME/bin:$HOME/.pub-cache/bin"
+dart pub global activate protoc_plugin
+
+echo "$LOG_PREFIX Build app"
+cd "$ROOT_DIR" && make
```
d-uzlov, line 40: In the last build I see that `bazel build` took 31 minutes of total 40 minutes for the whole `ci_post_clone.sh` script in the `Test - iOS` action.

Do I understand it correctly that we don't use any bazel cache in xcode cloud?

I see that "Clean environment" checkbox is not checked. Is it not applicable to bazel cache?
Can we move bazel cache folder to somewhere where xcode cloud will not delete it?

Can we connect xcode cloud to our Google Cloud Storage and use it for bazel remote caching?

flutter/ios/ci_scripts/ci_post_clone.sh
```diff
@@ -0,0 +1,45 @@
...
+cd "$ROOT_DIR"/flutter && flutter precache --ios
+cd "$ROOT_DIR"/flutter/ios && pod install
+cd "$ROOT_DIR"/flutter && flutter build ios --config-only integration_test/first_test.dart
+
+exit 0
```
d-uzlov, line 45: Please, add an empty line at the end of the file.

d-uzlov (changes requested):
Do I understand it correctly that xcode cloud doesn't allow us to reuse

d-uzlov:
Do I understand it correctly that xcode cloud doesn't allow us to reuse environment for testing and building a release, and we have to set up everything from scratch for each action (`Test` and `Archive`)?

What will happen if the test fails? Will we still upload the build to testflight for the corresponding commit in the current setup?

flutter/ios/ci_scripts/ci_post_clone.sh
d-uzlov: Maybe it would be possible to build the `ios_backend_fw_static.framework` in one action, upload it into Cloud Storage using commit hash and download it in other action(s) instead of building it?

flutter/ios/ci_scripts/ci_post_clone.sh
anhappdev: Ok

flutter/ios/ci_scripts/ci_post_clone.sh
anhappdev: Xcode Cloud should cache the data during the build but I did not find a documentation specified which data is cached. I will look into using Google Cloud Storage for bazel cache.

flutter/ios/ci_scripts/ci_post_clone.sh
d-uzlov: `Archive` action uploads resulting binary to testflight.
Since both `Test` and `Archive` actions use the same `ci_post_clone.sh` script, `Archive` uploads the binary that was built for testing.
If you install teslflight on your iphone and install `2.0.0 (17)` build from there, it will run benchmarks automatically, and will not allow you to interact with UI.

anhappdev:
> What will happen if the test fails? Will we still upload the build to testflight for the corresponding commit in the current setup?

There is a checkbox `Required to Pass` in the `Test` step, so I assume if the test fails, it will not upload the build in the next step `Archive`. I will test this behavior later.

flutter/ios/ci_scripts/ci_post_clone.sh
anhappdev: Ok, I will fix that, thanks for the catch.

d-uzlov:
By the way, I think we should set the `official-build=true` flag for testflight uploads.

anhappdev:
It would take some time to make the script using cached data. I deactivated the Xcode workflow for now, so we can still build manually when needed. I will update the script to cache data in another PR.

d-uzlov (approved):
Last automatic build in testflight still uses unofficial color scheme.
But we can probably think about it later.

d-uzlov:
It also seems like we won't really be able to use the local cache because it prevents automatic builds from being pushed to external testers.

Issue #354, opened 2022-06-08
App crashes when there are breaking changes in tasks file


d-uzlov:
After a recent commit that changed tasks file structure I got an error on app start:
<img src="https://user-images.githubusercontent.com/36223296/172705785-85b985d1-0106-44e4-a1b4-5c9d7d9cd474.PNG" width=300>

This happened because tasks file was already cached, and it didn't match the new structure.

We should avoid such tasks file updates in the future, we should change file name when there are breaking changes.
But the app should handle such errors gracefully.

I think we need the following changes:
1. Try redownloading tasks file if cached version failed to load
2. Show proper error message and allow user to select a different tasks file if redownloaded file still failed to load

anhappdev:
The easiest solution might be always remembering to update the task name but it's difficult to enforce.
I prefer to just be able to update the content of the file so we can easily use Git to see what changed.

We can add a checksum validation, the app validates the checksum every time it reads the file and redownloads if the checksum doesn't match.

d-uzlov:
I also thought about checksum validation.
The issues is that we don't want to pin the exact tasks file content.
I remember we discussed this at the end of last year. I suggested to point the default tasks file to the exact commit instead of a branch where it can be updated. But we decided that we want to keep the ability to change file contents dynamically without releasing a new version of the app.
Checksum validation would have the same issues as using commits instead of branches.

d-uzlov:
By the way, this also broke all of the old builds.
I guess we don't really care about automatic CI builds I think we do care about TestFlight builds.
The last build in testflight is broken on fresh install because the structure has changed.
But if I release a new build it will be broken on non-fresh installs because there is the old tasks file in the cache.

@anhappdev, I suggest to create a new PR in `mobile_models` repo that will revert changes and create a new file with updated structure, and then update the app to point to new link.

d-uzlov:
> The easiest solution might be always remembering to update the task name but it's difficult to enforce.

I think we should be able to set up CI to check this.
We can name tasks files `flutterapp_<commit>.pbtxt`, where `<commit>` is the corresponding commit of the `/flutter/cpp/proto/mlperf_task.proto` file.
In the `mobile_models` repo we can checkout `mobile_open` repo at corresponding commit and check that we are able to parse corresponding tasks file.

If someone changes structure of the tasks file the CI will fail, and they will have to create a new file with new name.

This way we will introduce an additional link between repositories but I don't this would be an issue.

[...]

PR #372, opened 2022-06-20
Handle errors in local task config

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/354

When current config is corrupted, app shows the following screen which allows user to get the app to a working state:


d-uzlov:
@anhappdev this PR is not intended to fix issues with default task config, it's for custom configs.

We have no control over custom configs, so moving the default config anywhere wouldn't help here.

flutter/lib/ui/settings/benchmarks_configuration_screen.dart
```diff
@@ -5,16 +5,17 @@ import 'package:provider/provider.dart';
 import 'package:mlperfbench/benchmark/state.dart';
 import 'package:mlperfbench/localizations/app_localizations.dart';
 import 'package:mlperfbench/resources/config_manager.dart';
+import 'package:mlperfbench/store.dart';
 import 'package:mlperfbench/ui/error_dialog.dart';
 
 class BenchmarksConfigurationScreen extends StatelessWidget {
-  final List<BenchmarksConfig> _benchmarksConfigurations;
+  final List<TaskConfigDescription> _benchmarksConfigurations;
```
anhappdev, line 12: Please update all the `benchmarksConfigurations` string in the repo to match with the new name `TaskConfigDescription`.
BTW, I think `TaskFile` describes the class more accurate.

anhappdev (changes requested):

flutter/lib/ui/settings/benchmarks_configuration_screen.dart
d-uzlov: Done.

I have renamed everything to `TaskConfig` / `Task configuration`.
This class is `...Description` because it contains filepath _and_ name.

I didn't change `benchmarksConfigurations.json` file name to avoid confusing users who already know about this file, and to avoid breaking it for users who already use this file.
I think in future we should add UI that would allow users to edit the list of configs without manually editing files, then filename will be irrelevant.

anhappdev: This one also needs to be updated:
https://github.com/mlcommons/mobile_app_open/blob/903bd43c8d2faf5eff0e9104e00a5531f1c11336/flutter/documentation/custom-tasks.md?plain=1#L37

> I think in future we should add UI that would allow users to edit the list of configs without manually editing files, then filename will be irrelevant.

Please create an issue for it so we can prioritize it.

d-uzlov: >This one also needs to be updated:

Done

>Please create an issue for it so we can prioritize it.

Done
https://github.com/mlcommons/mobile_app_open/issues/383

anhappdev (approved):

PR #411, opened 2022-06-28
Change pixel backend name as "TFLite-pixel"

kcwu:
So it could be distinguished with the default TFLite backend.


kcwu:
@freedomtan @ptt2panda PTAL a minor change

freedomtan (approved):

Issue #307, opened 2022-05-16
App crashes without showing `ExceptionWidget`


anhappdev:
The app should show the `ExceptionWidget` (a screen to display crash messages) when it throws an exception, but it's not.
We need to check which functions are impacted and handle it. (see https://github.com/mlcommons/mobile_app_open/pull/305#discussion_r873418883)

List of affected functions (last updated on 2022-06-15):
* https://github.com/mlcommons/mobile_app_open/blob/193b423222ad6def7dabb2adcef78ec762648cec/flutter/lib/resources/resource_manager.dart#L82
* https://github.com/mlcommons/mobile_app_open/blob/a827e3f2548274a875b4c0614f3fc13163c2fc06/flutter/lib/benchmark/state.dart#L207

anhappdev:
@d-uzlov Please take over this issue if you find the time for it.

PR #414, opened 2022-06-29
Properly handle exceptions in handleResources and runBenchmarks

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/307

If some error happens when running benchmarks application shows popup with error.
If some error happens during resource download (for example, checksum validation fails) application shows `ResourceErrorScreen` with "Clear cache" button.


anhappdev (changes requested):
On the first start after a clean install, the progress screen for loading resource did not show anymore, but only a black screen, until all the resources are downloaded.

anhappdev:
Please also add a log message in 
https://github.com/mlcommons/mobile_app_open/blob/03970131e87bab1d04ead6bad495718eb09e067e/flutter/lib/resources/file_cache_helper.dart#L46
so we can see it is still running when downloading large files, e.g.:

```dart
  Future<File> _download(String url) async {
    print('Downloading $url');
    const succesStatusCode = 200;
...
```

d-uzlov:
@anhappdev, done.

anhappdev (approved):

PR #425, opened 2022-07-07
Adding docker for DLC generation

mohitmundhragithub:
The previous docker file was removed from the flutter version of the app. So, adding it back, to be used for DLC generation for QTI backend.


d-uzlov (dismissed):

anhappdev (changes requested):
Can you move the file `flutter/android/docker/dataset_models/Dockerfile` to `datasets/Dockerfile`? I think it's better organized that way.

d-uzlov:
> Can you move the file `flutter/android/docker/dataset_models/Dockerfile` to `datasets/Dockerfile`?

I think it should be `datasets/docker/Dockerfile`, for consistency with other dockerfiles.

mohitmundhragithub:
> > Can you move the file `flutter/android/docker/dataset_models/Dockerfile` to `datasets/Dockerfile`?
> 
> I think it should be `datasets/docker/Dockerfile`, for consistency with other dockerfiles.

the dockerfile is required to be used for model generation as well. So, thought of keeping it in flutter/android.

mohitmundhragithub:
updated the paths

anhappdev:
> the dockerfile is required to be used for model generation as well. So, thought of keeping it in flutter/android.

I did not test it, but would be great if we can use the already existed `flutter/android/docker/Dockerfile` instead of the newly uploaded `datasets/docker/Dockerfile`.

anhappdev (approved):

Issue #383, opened 2022-06-22
Add UI to edit task config list


d-uzlov:
Currently the app can switch between several task config files in settings but the list must be created by manually editing a `json` file in application directory.

We should create UI controls to add/remove custom configs without manually editing files.

anhappdev:
One issue with using a user editable configuration file is that if we want to update the path to the default task file, it will not work if the users still have the old configuration file. So at least the path to the default task file should not be read from the configuration file. Or if we have the UI for editing the list, the configuration file should not be accessible / editable by the user.

anhappdev:
> if we want to update the path to the default task file, it will not work if the users still have the old configuration file

It seems to be a bug for me, so I created an issue to fix it first https://github.com/mlcommons/mobile_app_open/issues/420

PR #427, opened 2022-07-07
Adding the capability to use local datasets by avoiding download

mohitmundhragithub:
steps to follow,

1) build the app with above PR. 
2) Make sure the model files are also set to use "app:///mlperf_models" paths. 
3) if there is internet connectivity on the device, then the app will download the datasets, from github
4) if there is no internet connectivity on the device, then it will show an error screen.
5) tap on "select task config" link on the above error screen.
6) select "offline" task config from the task config window.
7) make sure the lite_datasets and mlperf_models folders are in "app:///" directory.
8) click on "Go"


flutter/lib/resources/config_manager.dart
```diff
@@ -12,6 +12,8 @@ import 'utils.dart';
 const _configListFileName = 'benchmarksConfigurations.json';
 const _defaultConfigName = 'default';
 const _defaultConfigUrl = 'asset://assets/tasks.pbtxt';
+const _offlineConfigName = 'offline';
+const _offlineConfigUrl = 'asset://assets/tasks_offline.pbtxt';
```
anhappdev, line 16: Adding a new duplicated task file with only some modified path will make it harder to maintain  it. Since the app has the ability to use custom task file, it should be preferred.

anhappdev (changes requested):
Maybe we can add an internet connectivity check before loading resources and if there is no internet connection, the app will show an error screen with the option to choose other task file.

mohitmundhragithub:
Yeah. That is what is happening with this app.
I have modified one of the option to show the task config file page, if internet resources are not available.

d-uzlov:
I think the ability to choose a custom task config from the error screen is useful.
But I'm hesitant about adding a mostly duplicate task config file. Also, current code wasn't really written to handle several default configs.

mohitmundhragithub:
> I think the ability to choose a custom task config from the error screen is useful. But I'm hesitant about adding a mostly duplicate task config file. Also, current code wasn't really written to handle several default configs.

we still have a single default config. offline config has been added in the app directly, instead of the user adding it.

Do you suggest any other way of doing it?

flutter/lib/resources/config_manager.dart
mohitmundhragithub: any suggestion on how to go about it? The process of using a custom task config file is quite cumbersome.

anhappdev: Yes, it's not convenient but it's one-time work to add a custom task file. We also have other issue to improve the UX of it (https://github.com/mlcommons/mobile_app_open/issues/383).

For this PR, I think just adding the option to chose other task file should be sufficient.

anhappdev: To be concrete, I mean just adding a new button `Choose other config` which then show the `TaskConfigScreen` should be sufficient. No need to add the `tasks_offline.pbtxt`.

d-uzlov:
> > I think the ability to choose a custom task config from the error screen is useful.
> > But I'm hesitant about adding a mostly duplicate task config file. Also, current code wasn't really written to handle several default configs.
> 
> we still have a single default config. offline config has been added in the app directly, instead of the user adding it.
> 
> Do you suggest any other way of doing it?

I just think that we need a refactoring in config loading to accommodate for several default configs.
But it's a small issue, we could fix it later.

My primary concern is that we will need to support a second task config file. It will not be covered by testing on CI, it would be easy to make a mistake when editing configs. We will have to test it manually.
And I'm not sure if it would be useful for many people.
I agree with Anh that this seems to be a case for using a custom config.

mohitmundhragithub:
sure, i will push the changes based on above discussion. Thank you.

d-uzlov:
I suggest we discuss the need for a second default task config in the next weekly meeting.
I think for now it would be better to remove the second config from this PR, merge the changes for the Error Screen.
And if we decide to add the second config, I'll do it in a separate PR with proper refactoring.

mohitmundhragithub:
> I suggest we discuss the need for a second default task config in the next weekly meeting. I think for now it would be better to remove the second config from this PR, merge the changes for the Error Screen. And if we decide to add the second config, I'll do it in a separate PR with proper refactoring.

This won't be needed. I figured out a way to use the tasks_offline. I think we can completely avoid adding that task config file.

I have uploaded the changes as discussed. Please have a look again. Thank you for all the suggestions.

mohitmundhragithub:
I see format and lint error for one of the file. 
how to correct the formatting? I know formatter is part of the repo, but can someone please share the steps to run it?

anhappdev:
> I see format and lint error for one of the file.
> how to correct the formatting? I know formatter is part of the repo, but can someone please share the steps to run it?

`make docker/format`
https://github.com/mlcommons/mobile_app_open/blob/master/CONTRIBUTING.md#coding-styles

anhappdev (approved):

Issue #423, opened 2022-07-04
Link backend dynamically on iOS


d-uzlov:
Currently we use static linking on iOS.
We should look into using dynamic libraries for more flexibility.

PR #430, opened 2022-07-08
Use dynamic linking for iOS native libs

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/423


flutter/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -32,12 +35,14 @@
...
 /* Begin PBXCopyFilesBuildPhase section */
-		9705A1C41CF9048500538489 /* Embed Frameworks */ = {
+		19C134552876922700263A48 /* Embed Frameworks */ = {
 			isa = PBXCopyFilesBuildPhase;
 			buildActionMask = 2147483647;
 			dstPath = "";
 			dstSubfolderSpec = 10;
 			files = (
+				19C134542876922700263A48 /* backend_bridge_fw.xcframework in Embed Frameworks */,
+				19C1345828769D8300263A48 /* tflite_backend_fw.xcframework in Embed Frameworks */,
```
d-uzlov, line 45: One issue I wasn't able to solve is that XCode tries to bundle all listed frameworks. If we were to disable TFLite backend we would get a compile error about missing framework.
We will need to create an empty dummy backend to include it instead of the real one if we want to be able to disable it (without using xcode UI).

d-uzlov, line 45: Xcode links all of the bundled libraries automatically.
I tried adding second instance of TFLite backend to the app. Xcode emitted a warning about duplicated symbols but there were no errors.
I think it shouldn't because we load libraries an its exported functions manually, and stuff inside libraries should already be linked to itself. But I'm not 100% sure.

mobile_back_tflite/cpp/backend_tflite/ios/BUILD
```diff
@@ -0,0 +1,22 @@
+load("@build_bazel_rules_apple//apple:apple.bzl", "apple_xcframework")
+
+apple_xcframework(
+    name = "libtflitebackend",
```
d-uzlov, line 4: I created a separate file for apple tflite target to ensure that there are no name conflicts.

flutter/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -3,22 +3,25 @@
...
-		035547E55D1857EA2B3C8779 /* Pods_RunnerTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 10CDB3E2A91DF927CD274CD3 /* Pods_RunnerTests.framework */; };
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
+		19C134532876922700263A48 /* backend_bridge_fw.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 19C1344F2876919E00263A48 /* backend_bridge_fw.xcframework */; };
+		19C134542876922700263A48 /* backend_bridge_fw.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 19C1344F2876919E00263A48 /* backend_bridge_fw.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		19C1345E2878151A00263A48 /* libtflitebackend.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 19C1345D2878151A00263A48 /* libtflitebackend.xcframework */; };
+		19C1345F2878151A00263A48 /* libtflitebackend.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 19C1345D2878151A00263A48 /* libtflitebackend.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
d-uzlov, line 16: One issue I wasn't able to solve is that XCode tries to bundle all listed frameworks. If we were to disable TFLite backend we would get a compile error about missing framework.
We will need to create an empty dummy backend to include it instead of the real one if we want to be able to disable it (without using xcode UI).

d-uzlov, line 16: Xcode links all of the bundled libraries automatically.
I tried adding second instance of TFLite backend to the app. Xcode emitted a warning about duplicated symbols but there were no errors.
I think it shouldn't be an issue because we load libraries and its exported functions manually, and stuff inside libraries should already be linked to itself.

.bazelrc
```diff
@@ -43,8 +43,6 @@ build:android_x86_64 --fat_apk_cpu=x86_64
 build:ios --apple_platform_type=ios
 build:ios --apple_bitcode=embedded --copt=-fembed-bitcode
 build:ios --copt=-Wno-c++11-narrowing
-build:ios_fat64 --config=ios
-build:ios_fat64 --ios_multi_cpus=arm64,x86_64
 
 # Fix an error in build_bazel_rules_apple
 build:ios --incompatible_run_shell_command_string=false
```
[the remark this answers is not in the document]
d-uzlov: Done

mobile_back_tflite/cpp/backend_tflite/BUILD
```diff
@@ -124,7 +124,6 @@ tflite_jni_binary(
 
 tflite_jni_binary(
     name = "libtfliteneuronbackend.so",
-    exported_symbols = "//flutter/cpp/c:exported_symbols.lds",
```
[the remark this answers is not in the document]
d-uzlov: Good catch.
`exported_symbols` is only used by xcode compiler so we don't need it on android targets.
I have removed it from all targets from all backends.

anhappdev (approved):

mobile_back_tflite/cpp/backend_tflite/BUILD
```diff
@@ -60,82 +60,14 @@ cc_library(
...
-            "@org_tensorflow//tensorflow/lite/delegates/gpu:delegate",
-        ],
-        "//conditions:default": [],
-    }),
-    alwayslink = 1,
-)
-
 tflite_jni_binary(
     name = "libtflitebackend.so",
-    exported_symbols = "//flutter/cpp/c:exported_symbols.lds",
```
anhappdev, line 118: @d-uzlov Removing this caused an `symbol not found` error when testing the backend via command line (https://github.com/mlcommons/mobile_app_open/blob/master/flutter/cpp/binary/README.md)
You can also clone this branch `360-add-coreml-backend` and run `cd mobile_back_apple && make tflite` to test.

freedomtan: fortunately, this should be a macOS/iOs specific problem because macOS/iOS's linker (ld) doesn't support linker script.

freedomtan: we can find how `tflite_jni_binary()` handles this at https://github.com/tensorflow/tensorflow/blob/r2.9/tensorflow/lite/build_def.bzl#L154-L190

for macOS, a exported symbol list, `-Wl,-exported_symbols_list,...`, is used
for others, linker script, `-Wl,--version-script,...`, is used

anhappdev: @freedomtan 👍

PR #436, opened 2022-07-14
use latest v2.1 loadgen

### LINKED ISSUE #435: Use "latest" loadgen for v2.1 submission
There is newer loadgen from the repo of the inference group, https://github.com/mlcommons/inference/tree/r2.1. Let's update our 2.1 branch to use it.

freedomtan:
this builds well, but doesn't work because the removal of [ReportLatencyResults() in loadgen ](https://github.com/mlcommons/inference/commit/a8a43dcf2ce9f795be637f3126250c1cfa0d736e)

close #435

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -90,10 +90,10 @@ http_archive(
     name = "org_mlperf_inference",
     build_file = "@//flutter/android/third_party:loadgen.BUILD",
     patch_cmds = ["python3 loadgen/version_generator.py loadgen/version_generated.cc loadgen"],
-    sha256 = "f4c57a3f3cd71f2dac166a79ad760b824aafda7b91400889acff4a9c7dbdaf8e",
-    strip_prefix = "inference-a77ac37d07145d9f3123465a8fd18f9ebbde5d6a",
+    sha256 = "6773b97450d3cfe153affd8e0c69ad2c85393b1867712e812ea0d32336fa7041",
+    strip_prefix = "inference-5dd9ed739cc8dfeae2c5d30ec06f4cb524a4dd74",
     urls = [
-        "https://github.com/mlcommons/inference/archive/a77ac37d07145d9f3123465a8fd18f9ebbde5d6a.tar.gz",
+        "https://github.com/mlcommons/inference/archive/5dd9ed739cc8dfeae2c5d30ec06f4cb524a4dd74.tar.gz",
     ],
 )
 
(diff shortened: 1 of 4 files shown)


freedomtan:
@d-uzlov could you help on this? I guess we have to avoid items depending on result from LoadGen's ReportLatencyResults()

flutter/cpp/mlperf_driver.h
```diff
@@ -56,13 +56,6 @@ class MlperfDriver : public ::mlperf::SystemUnderTest {
   // Flush the staged queries immediately.
   void FlushQueries() override { backend_->FlushQueries(); }
 
-  // Called by loadgen to show us the recorded latencies.
-  void ReportLatencyResults(
-      const std::vector<::mlperf::QuerySampleLatency>& latencies_ns) final {
-    latencies_ns_.insert(latencies_ns_.end(), latencies_ns.begin(),
```
d-uzlov, line 62: `latencies_ns_` is used to calculate performance metric:
https://github.com/mlcommons/mobile_app_open/blob/c7dac5e560674652b28a1ad768099211c3cb105d/flutter/cpp/mlperf_driver.h#L67-L68
When it is empty, app UI gets `0.0` latency for all results, which results in `infinity` performance value, and this causes the error that we can see on the screenshot.

We can either create a patch for loadgen to revert the commit that removes `ReportLatencyResults`, or we can parse loadgen summary logs.

It would be ideal if loadgen exported summary in its API instead of only printing it to file but I guess we won't get this anytime soon.

freedomtan: Thanks, I created a patch to revert LoadGen's removal. It works.

WORKSPACE
```diff
@@ -89,11 +89,15 @@ new_local_repository(
 http_archive(
     name = "org_mlperf_inference",
     build_file = "@//flutter/android/third_party:loadgen.BUILD",
+    patch_args = ["-p1"],
     patch_cmds = ["python3 loadgen/version_generator.py loadgen/version_generated.cc loadgen"],
-    sha256 = "f4c57a3f3cd71f2dac166a79ad760b824aafda7b91400889acff4a9c7dbdaf8e",
-    strip_prefix = "inference-a77ac37d07145d9f3123465a8fd18f9ebbde5d6a",
+    patches = [
+        "//patches:revert_RAeportLatencyResults.patch",
```
d-uzlov, line 95: Build fails because of the typo in patch filename.
It should be `revert_ReportLatencyResults`.

d-uzlov (approved):

Issue #432, opened 2022-07-11
Disable Java linter


d-uzlov:
We no longer use Java in this repo (in the master branch) after removal of the old Android app.
We should disable Java formatter/linter.

PR #440, opened 2022-07-14
Disable java linter

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/432


anhappdev (changes requested):
The file `datasets/docker/Dockerfile` is used by the vendor to generate dataset and model files. I would leave it untouched.

anhappdev (approved):

Issue #456, opened 2022-07-27
Check accuracy on CI


d-uzlov:
Currently on CI we don't check accuracy at all so we may miss changes that don't immediately crash the app.

Steps:

- Add groundtruth files to test dataset
    - [x] Imagenet
    - [x] COCO
    - [x] ADE20K
- [x] Test and save current accuracy (per each tested backend)
- [x] Check that accuracy match the saved values in the integration test

Parent issue: https://github.com/mlcommons/mobile_app_open/issues/347

PR #473, opened 2022-08-04
Check accuracy on CI

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/456

### DIFF
--- a/flutter/cpp/datasets/ade20k.cc
+++ b/flutter/cpp/datasets/ade20k.cc
@@ -213,7 +213,14 @@ float ADE20K::ComputeAccuracy() {
   }
   float iou_sum = 0.0;
   for (int j = 0; j < num_classes_; j++) {
-    auto iou = tp_acc_[j] * 1.0 / (tp_acc_[j] + fp_acc_[j] + fn_acc_[j]);
+    auto sum = tp_acc_[j] + fp_acc_[j] + fn_acc_[j];
+    if (sum == 0) {
+      // in our integration test we use very small dataset
+      // which doesn't have some of the classes
+      iou_sum += 1.0;
+      continue;
+    }
+    auto iou = tp_acc_[j] * 1.0 / sum;
 #if __DEBUG__
     LOG(INFO) << "IOU class " << j + 1 << ": " << tp_acc_[j] << ", "
               << fp_acc_[j] << ", " << fn_acc_[j] << ", " << iou << "\n";
--- a/flutter/cpp/datasets/ade20k.cc
+++ b/flutter/cpp/datasets/ade20k.cc
@@ -217,7 +217,6 @@ float ADE20K::ComputeAccuracy() {
     if (sum == 0) {
       // in our integration test we use very small dataset
       // which doesn't have some of the classes
-      iou_sum += 1.0;
       continue;
     }
     auto iou = tp_acc_[j] * 1.0 / sum;
--- a/flutter/integration_test/expected_accuracy.dart
+++ b/flutter/integration_test/expected_accuracy.dart
@@ -37,13 +37,9 @@ const Map<String, Interval> _languageUnderstanding = {
 };
 
 const benchmarkExpectedAccuracy = {
-  'IC_tpu_float32': _imageClassification,
-  'IC_tpu_uint8': _imageClassification,
-  'OD_float32': _objectDetection,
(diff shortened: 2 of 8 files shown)


d-uzlov:
Currently this PR uses datasets from my branch in the mobile_models repo.
I'll update dataset links after we merge https://github.com/mlcommons/mobile_models/pull/47.

flutter/integration_test/first_test.dart
```diff
@@ -70,16 +75,40 @@ void main() {
...
+            expectedAccuracy[backendFilename]![benchmarkResult.benchmarkId];
+        expect(
+          expectedAccuracyValue,
+          isNotNull,
+          reason:
+              'missing accuracy expected value for $backendFilename[${benchmarkResult.benchmarkId}]',
+        );
+        expect(
+          benchmarkResult.accuracy!.accuracy!.normalized,
+          greaterThanOrEqualTo(expectedAccuracyValue!),
```
anhappdev, line 109: I would `expect` the accuracy to be in a range instead of `greaterThanOrEqualTo`. 
So maybe this in pseudo code:
```
tolerance  = 0.05
expect(realValue, greaterThanOrEqualTo(expectedValue * (1-tolerance)))
expect(realValue, lessThanOrEqualTo(expectedValue * (1+tolerance)))
```

flutter/cpp/datasets/ade20k.cc
```diff
@@ -213,7 +213,14 @@ float ADE20K::ComputeAccuracy() {
...
   for (int j = 0; j < num_classes_; j++) {
-    auto iou = tp_acc_[j] * 1.0 / (tp_acc_[j] + fp_acc_[j] + fn_acc_[j]);
+    auto sum = tp_acc_[j] + fp_acc_[j] + fn_acc_[j];
+    if (sum == 0) {
+      // in our integration test we use very small dataset
+      // which doesn't have some of the classes
+      iou_sum += 1.0;
+      continue;
+    }
+    auto iou = tp_acc_[j] * 1.0 / sum;
```
anhappdev, line 223: @freedomtan Please check if this change has an impact on the submission, since the model / backend can achieve higher accuracy by just ignoring some classes.

anhappdev (changes requested):

flutter/cpp/datasets/ade20k.cc
d-uzlov: Backends can't use this because the `if (sum == 0)` branch is triggered only when _dataset_ doesn't have some classes.
And this change doesn't really affect us because if you mess with datasets in an appropriate way, you can reach arbitrary high accuracy, regardless of this change.

Still I updated the code to make accuracy lower when dataset doesn't contain some classes, instead of adding `+1` for missing classes.

flutter/integration_test/first_test.dart
d-uzlov: Done.
I have made it so you can specify expected range. Some benchmarks always produce the exact same result, and some have variations of different size.

flutter/cpp/datasets/ade20k.cc
anhappdev: Since the updated code doesn't increase accuracy anymore, it should be ok now.

flutter/integration_test/expected_accuracy.dart
```diff
@@ -0,0 +1,50 @@
+class Interval {
+  final double min;
+  final double max;
+
+  const Interval({required this.min, required this.max});
+}
+
+const Map<String, Interval> _imageClassification = {
+  'CPU': Interval(min: 1.0, max: 1.0),
+  'NNAPI': Interval(min: 0.8999999761581421, max: 0.8999999761581421),
```
anhappdev, line 10: The number can just have only 2 or 3 decimal places to make it easy to read, since we use an interval right now.

d-uzlov: Done.
I reduces the precision to 4 decimal places since this is what we already use in the UI.

anhappdev (approved):

freedomtan (dismissed):

d-uzlov:
@freedomtan, could you approve https://github.com/mlcommons/mobile_models/pull/47, please?
We should merge that PR before this one.

anhappdev (approved):

PR #481, opened 2022-08-18
Improve Core ML performance

### LINKED ISSUE #472: Improve performance of Core ML backend
We have just implemented the Core ML backend, the focus was to make it work first, then make it work faster.
The performance of the Core ML backend should be equal or better than TFLite backend.

anhappdev:
Closes #472


freedomtan:
close this. Since we can most confirm that it's mostly scheduler + dvfs issue. And we have "artificial load" as an option already.

PR #493, opened 2022-09-02
Add SonarScanner

### LINKED ISSUE #488: Add automated code check to CI
We should integrate this repo with a tool / service to auto check for code quality and code security.

This one is free for open-source project:
https://sonarcloud.io

If anyone has experience with other tools, please share.

anhappdev:
Closes #488


.github/workflows/code-analysis.yml
```diff
@@ -0,0 +1,37 @@
...
+      contents: read
+      security-events: write
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Build scanner image
+        run: |
+          make scanner/image
+      - name: Build app and scan code
```
d-uzlov, line 25: It would be more convenient to have separated build and scan stages.
We would know how much time each logical step takes, and if anything fails it would be easier to search for the cause.

tools/scanner/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM mlcommons/mlperf_mobile_flutter
+
+RUN apt-get update >/dev/null && DEBIAN_FRONTEND=noninteractive apt-get install -y unzip
+
+ARG SONAR_SCANNER_DIR=$HOME/.sonar
+
+# Install sonar-scanner
```
d-uzlov, line 7: Why do we need a custom image for this?
Aren't there some ready-to-use solutions?

For example, I was able to quickly find the following github action:
https://github.com/marketplace/actions/sonarqube-scanner

tools/scanner/scan.mk
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+# Note:
+# SonarScanner requires a clean build for every run. Do not cache build files.
```
d-uzlov, line 17: Why do we need a clean build?
Why do we even need a build for static analysis tool?

The issue I have with this is time.
Currently our CI is usually 30-35 minutes long, and I think it's already really slow (I wish we could make Windows CI faster).
From what I see in this PR checks, Sonar scan takes 50 minutes, roughly 1.5 times longer.
I think it's unacceptably long for a pull request check.

Even worse is that local checks will also take a long time.

tools/scanner/scan.mk
```diff
@@ -0,0 +1,66 @@
...
+.PHONY: scanner/build
+scanner/build:
+	bazel clean
+	build-wrapper-linux-x86-64 --out-dir "${SONAR_OUT_DIR}" \
+		make flutter/android
+  
+# TODO (anhappdev): Use MLCommons organization
+.PHONY: scanner/scan
+scanner/scan:
+	sonar-scanner \
```
d-uzlov, line 48: What code issues do we scan for?
How useful is it?
Since the CI passed without any changes to our current codebase I suspect that there are not many checks?

d-uzlov (changes requested):

tools/scanner/scan.mk
d-uzlov: If it's impossible to reduce time for `build + sonar checks`, we can just disable it for pull requests but keep for the master branch commits.

.github/workflows/code-analysis.yml
```diff
@@ -0,0 +1,37 @@
...
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_OUT_DIR: sonar_build_wrapper_output
+          PR_NUMBER: ${{ github.event.number }}
+          PR_BRANCH: ${{ github.HEAD_REF }}
+          PR_BASE: ${{ github.BASE_REF }}
+          CONNECT_TTY_TO_DOCKER: 0
+          WITH_TFLITE: 1
+          WITH_PIXEL: 1
+        run: |
+          make docker/scan
```
d-uzlov, line 37: Aren't there some limits of how much CPU time we can use in our github actions?
Can we run our of them if we use github to build our C++ code?

tools/scanner/Dockerfile
anhappdev: > For example, I was able to quickly find the following github action:
https://github.com/marketplace/actions/sonarqube-scanner

With only 2 star and last updated on 8 Jun 2020, I prefer not to use it.

BTW, here is the official one: https://github.com/SonarSource/sonarcloud-github-action
It does not work for compiled languages, though.

tools/scanner/scan.mk
anhappdev: > Why do we need a clean build?
Why do we even need a build for static analysis tool?

It stated in the installation guide.

> The issue I have with this is time.
>  just disable it for pull requests but keep for the master branch commits.

Maybe we can make it optional and not required to merge, but I prefer to let it run on PR so I can check and fix the code if I'd like to.

tools/scanner/scan.mk
anhappdev: > What code issues do we scan for?
How useful is it?

Please see https://github.com/mlcommons/mobile_app_open/issues/488#issuecomment-1236634456
and also their docs: https://docs.sonarcloud.io/improving/overview/

Currently, it does not support `dart`, but there is a 3rd party extension, maybe we can try it later.

.github/workflows/code-analysis.yml
anhappdev: The GitHub action is free for open-source repo. Certainly, there is some limitation or quota but I couldn't find it on the internet.

d-uzlov (approved):

freedomtan (approved):

PR #494, opened 2022-09-04
Add filtering by OS type to online results

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/505

This is kind of a proof of concept for filtering.
I'll create separate issues to add other options for filtering.
We may need to make changes to our data structure to support some of the filters.

Depends on:
* https://github.com/mlcommons/mobile_app_open/pull/482

### DIFF
--- a/firebase_functions/functions/src/fetch.ts
+++ b/firebase_functions/functions/src/fetch.ts
@@ -13,31 +13,17 @@ function _makeBatchQuery(
     .limit(pageSize);
 }
 
-export function fetchFirst(
-  db: FirebaseFirestore.Firestore,
-): (request: Request, response: Response, next: NextFunction) => Promise<void> {
-  return async (request: Request, response: Response, _: NextFunction) => {
-    try {
-      const pageSizeString = request.headers['page-size'];
-      if (typeof pageSizeString != 'string') {
-        throw Error('page-size header is missing');
-      }
-      const pageSize = parseInt(pageSizeString);
-      console.log(`pageSize is ${pageSize}`);
-
-      const snapshot = await _makeBatchQuery(db, pageSize).get();
-      const result = [];
-      for (const document of snapshot.docs) {
-        result.push(document.data());
-      }
-      console.log(result);
-      response.status(StatusCodes.OK).send(JSON.stringify(result));
-    } catch (e) {
-      functions.logger.info(e, { structuredData: true });
-      response.status(StatusCodes.BAD_REQUEST).send('invalid request: ' + e);
-      return;
-    }
-  };
+function _getPageSize(headers: any): number {
+  const pageSizeString = headers['page-size'];
+  if (typeof pageSizeString != 'string') {
+    throw Error('page-size header is missing');
+  }
+  const maxAllowedPageSize = 50;
+  const result = parseInt(pageSizeString);
+  if (result > maxAllowedPageSize) {
+    throw Error('page size is too big');
+  }
+  return result;
 }
 
 export function fetchNext(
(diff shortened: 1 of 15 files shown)


freedomtan (approved):
LGTM. Question: what it I use macOS or iPadOS to test?

d-uzlov:
@freedomtan I suppose you mean how hard it would be to add support for more operating systems to this filter.

Due to very limited support from FIrebase, the list of supported OS values has fixed size.
Right now there is support for 6 possible values, so we can add support for 3 more operating systems without any issues.
https://github.com/mlcommons/mobile_app_open/blob/e2ab6b6d91e2bf8cc0ae09e73060e9425b10ecc1/firebase_functions/functions/src/indices.ts#L9-L13
If we need to add more, there would be issues. It would still be possible but we would need to create a migration script that would update all results that will have been uploaded by that moment and increase the size of the index.
We can increase the number of reserved values right now if we expect to add support for 3+ systems.

d-uzlov:
@anhappdev, could you also review it, please?

anhappdev:
@d-uzlov To test this I would need to set up a Firebase account first, right?
I think we should add the MLCommons's Firebase account to this repo. It's quite impractical for reviewer to set up the Firebase account for testing.

d-uzlov:
@anhappdev, do you have access to mlcommons firebase api key?
https://console.firebase.google.com/project/mobile-app-build-290400/settings/general/web:YTRkYTRjYmMtNjYyMC00MDdiLTllMDktNjI5MDM1OGE3YTJm

I'm hesitant about adding api keys into the repo, even though Firebase says that it's ok.

anhappdev:
I have access to the link, so I will use it to test for now.

I did not use Firebase for web before, but for iOS or Android, there would be a config file, which is straightforward to read by open the app package, so the config data is not a secret.
https://firebase.google.com/docs/android/setup#add-config-file

But to avoid API abuse, I think we need to set up something like that first:
https://firebase.google.com/docs/app-check

d-uzlov:
>I did not use Firebase for web before

By the way, for our purposes web/ios/andoid is just a human-readable tag. I tried creating other apps, and configs are the same. Maybe it will matter if we start using other firebase features.

We can look into app-check in the future.

anhappdev:
Can you update the docs here:
https://github.com/mlcommons/mobile_app_open/blob/master/flutter/documentation/firebase.md#linking-flutter-to-firebase

Got an error ```make: *** No rule to make target `flutter/firebase-config'.  Stop.```

d-uzlov:
I'll update docs.
The correct target is just `flutter/firebase`, without `-config`.

anhappdev:
Got some other error when trying to upload result:

If I use:
The `FIREBASE_FLUTTER_FUNCTIONS_URL` with `xxxxxx.cloudfunctions.net` then got this error
`flutter: error 302:`

If I use:
The `FIREBASE_FLUTTER_FUNCTIONS_URL` with `xxxxxx.firebaseio.com` from `databaseURL` field then got this error
`flutter: error 405: append .json to your request URI to use the REST API.`

Trying to view the online tab also did not work.
```
flutter: FormatException: Unexpected character (at character 1)
<!doctype html><html lang="en-US" dir="ltr"><head><base href="https://accou...
^
flutter: #0      _ChunkedJsonParser.fail (dart:convert-patch/convert_patch.dart:1405:5)
#1      _ChunkedJsonParser.parseNumber (dart:convert-patch/convert_patch.dart:1272:9)
#2      _ChunkedJsonParser.parse (dart:convert-patch/convert_patch.dart:937:22)
#3      _parseJson (dart:convert-patch/convert_patch.dart:40:10)
#4      JsonDecoder.convert (dart:convert/json.dart:612:36)
#5      JsonCodec.decode (dart:convert/json.dart:216:41)
#6      jsonDecode (dart:convert/json.dart:155:10)
#7      RestHelper.fetchFirst (package:mlperfbench_common/firebase/rest_helper.dart:56:22)
<asynchronous suspension>
#8      FirebaseCacheHelper.fetchBatch (package:mlperfbench_common/firebase/cache_helper.dart:23:16)
```

Tested on iPhone simulator.

d-uzlov:
@anhappdev are you trying to use real firebase servers or the firebase emulator?
Have you set up the corresponding instance of firebase functions?
https://github.com/mlcommons/mobile_app_open/blob/8c86121160afa131490c79244db9341d58f60ac1/flutter/documentation/firebase.md#developing-firebase-functions

anhappdev:
I only do the set-up in [Linking Flutter to Firebase](https://github.com/mlcommons/mobile_app_open/blob/8c86121160afa131490c79244db9341d58f60ac1/flutter/documentation/firebase.md#linking-flutter-to-firebase).

I now use the existed `functionsPrefix` `devd2208` in the project. Upload result works but still got error showing the online tab:
```
[VERBOSE-2:profiler_metrics_ios.mm(203)] Error retrieving thread information: (os/kern) invalid argument
flutter: error 400: invalid request: Error: Value for argument "documentPath" must point to a document, but was "v0/". Your path does not contain an even number of components.
flutter: #0      RestHelper.fetchNext (package:mlperfbench_common/firebase/rest_helper.dart:64:7)
<asynchronous suspension>
#1      FirebaseCacheHelper.fetchBatch (package:mlperfbench_common/firebase/cache_helper.dart:31:14)
<asynchronous suspension>
#2      OnlineTab.fetchData (package:mlperfbench/ui/history/online_tab.dart:130:7)
<asynchronous suspension>
```

d-uzlov:
@anhappdev, I uploaded an updated version into firebase but it doesn't work.
It seems like firebase requires us to set up some additional indices that aren't needed in emulator.
I'll figure out how to fix it and ping you.

d-uzlov:
We no longer need this PR because we decided to change the approach to filtering.
Maybe it would be possible to extract some UI changes from this PR and use it for the new approach, but not in this PR.

Issue #477, opened 2022-08-15
Meaning of `configuration` in backend settings


d-uzlov:
We have `configuration` field defined for each benchmark.

https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/flutter/cpp/proto/backend_setting.proto#L48-L49 https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h#L36 https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/mobile_back_samsung/samsung/lib/sbe_config.hpp#L34 https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/mobile_back_qti/cpp/backend_qti/qti_settings.h#L36 https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/mobile_back_apple/cpp/backend_coreml/coreml_settings.h#L27

What does this value actually mean? I can't really infer it from current config files, each backend uses it for different things.
Is it supposed to name underlying machine learning framework?

freedomtan:
is this for underlying accelerator configurations we discussed before?

freedomtan:
This is kinda what Qualcomm colleagues use to display the "backend" to use (sometime it's TFLite + GPU delegate, not always SNPE).

d-uzlov:
I'll create a PR to update documentation about this field.

PR #499, opened 2022-09-05
Update backend `configuration` field documentation

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/477

### DIFF
--- a/flutter/cpp/proto/backend_setting.proto
+++ b/flutter/cpp/proto/backend_setting.proto
@@ -44,9 +44,9 @@ message BenchmarkSetting {
   required string model_checksum = 4;
   // accelerator is passed to the backend
   required string accelerator = 5;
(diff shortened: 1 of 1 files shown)


mohitmundhragithub (approved):

mohitmundhragithub:
don't we need changes in the tasks.pbtxt file in this PR?

d-uzlov:
This PR doesn't change meaning of the `configuration` field, it just writes it in the documentation so you don't have to guess and ask other people about it.
Also, this field is a part of backend configuration, and tasks.pbtxt is the frontend config.

anhappdev (approved):
The name `configuration` does not describe accurately how we use it `// Human-readable name of the underlying ML framework`.
Maybe we should change from `configuration` to `framework`.

Issue #507, opened 2022-09-06
Convert model codename to human-readable name for online results


d-uzlov:
Currently we get and save model as the OS reports it.
OS actually reports a model identifier instead of human-readable model name.

We should look for a way to convert model id into model name.

We can probably use the following library:
https://pub.dev/packages/device_marketing_names

freedomtan:
this is for (android) device code name.

PR #510, opened 2022-09-07
Refactor environment info generation

d-uzlov:
Part of https://github.com/mlcommons/mobile_app_open/issues/507

### DIFF
--- a/flutter/lib/device_info.dart
+++ b/flutter/lib/device_info.dart
@@ -67,7 +67,7 @@ class DeviceInfo {
 
   static Future<DeviceInfo> _makeWindowsInfo() async {
     return DeviceInfo(
-      manufacturer: 'Microsoft',
+      manufacturer: '',
       modelCode: 'Unknown PC',
       modelName: '',
     );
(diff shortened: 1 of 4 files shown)


flutter/lib/device_info.dart
```diff
@@ -7,44 +7,80 @@ import 'package:mlperfbench_common/data/environment/environment_info.dart';
...
+    return DeviceInfo(
+      manufacturer: deviceInfo.manufacturer,
+      modelCode: deviceInfo.model,
+      modelName: '',
+    );
+  }
+
+  static Future<DeviceInfo> _makeWindowsInfo() async {
+    return DeviceInfo(
+      manufacturer: 'Microsoft',
```
anhappdev, line 70: Is `Microsoft` the manufacturer for all Windows devices?

anhappdev (changes requested):

flutter/lib/device_info.dart
d-uzlov: No.
I also caught this issue when working on #512 but forgot to push it into this PR.

anhappdev (approved):

PR #516, opened 2022-09-14
Add a setting to enable artificial CPU load in iOS

anhappdev:
A new toggle button is added to the Settings screen to enable the artificial load. The button is only visible on iOS platform.
This is a hack to keep CPU utilization high on Apple devices, which should enable P core to achieve high throughput.
On my iPhone XS and Mac M1, CPU load increases from ~30% to ~130% if enabled.


freedomtan:
Confirm app built with this PR works as expected on an iPhone 13.

anhappdev:
An iOS build based on this PR is now in TestFlight (`2.0.0 (42)`)

d-uzlov (approved):

PR #517, opened 2022-09-15
Optimize Code Analysis workflow

anhappdev:
Cancel previous run on new commit and some minor changes in the configuration.

### DIFF
--- a/tools/scanner/scan.mk
+++ b/tools/scanner/scan.mk
@@ -50,6 +50,7 @@ scanner/scan:
 		-Dsonar.organization=anhappdev \
 		-Dsonar.projectKey=mobile_app_open \
 		-Dsonar.sources=. \
+		-Dsonar.python.version="3.7, 3.8, 3.9" \
 		-Dsonar.cfamily.build-wrapper-output="${SONAR_OUT_DIR}" \
 		-Dsonar.host.url=https://sonarcloud.io \
 		-Dsonar.scm.provider=git \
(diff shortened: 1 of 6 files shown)


anhappdev:
Could not get the 3rd party Flutter plugin works so just wait for the official plugin.

tools/scanner/scan.mk
```diff
@@ -40,8 +41,7 @@ scanner/image: output/docker_mlperf_scanner.stamp
 .PHONY: scanner/build
 scanner/build:
 	bazel clean
-	build-wrapper-linux-x86-64 --out-dir "${SONAR_OUT_DIR}" \
-		make flutter/android
+	build-wrapper-linux-x86-64 --out-dir "${SONAR_OUT_DIR}" make
```
d-uzlov, line 44: `flutter/prepare` target which is implied here takes some time. I think it would be better to revert it to `make flutter/android` until we enable Flutter code analysis.

tools/scanner/scan.mk
```diff
@@ -58,7 +59,7 @@ scanner/scan:
 		-Dsonar.pullrequest.base=${PR_BASE}
 
 .PHONY: docker/scanner/build
-docker/scanner/build: scanner/image
+docker/scanner/build: output/docker_mlperf_scanner.stamp
```
d-uzlov, line 62: I thought the whole reason to have the `scanner/image` target is to avoid using the `.stamp` target. At the very least it's easier to read.
Am I missing something here?

tools/scanner/scan.mk
anhappdev: Done

tools/scanner/scan.mk
anhappdev: There is already a build image step before, so I though adding that `.stamp` would prevent the image from being built again if already happened, but somehow it didn't work. So I removed the `.stamp` dependency now.

d-uzlov (approved):

Issue #489, opened 2022-08-29
Change default time unit from `milliseconds` to `seconds`


d-uzlov:
In the config we use values of the order of seconds. It would be easier to read if we remove redundant `000`.
We can use floating point values to keep high precision.
https://github.com/mlcommons/mobile_app_open/blob/a67fb3125ff92cb1dca5b0b5cbf28eb90b482010/flutter/assets/tasks.pbtxt#L5
```yaml
min_duration: 60.0
```

Regarding local variables and JSON results there is no meaningful difference between seconds and milliseconds for the program.
But milliseconds require additional constant for many arithmetic operations which makes it harder to maintain.

Also `second` is the main time unit in the SI unit system.

Depends on:
* https://github.com/mlcommons/mobile_app_open/pull/484

anhappdev:
I think second is not precise enough for this `benchmarking` app. To convert between unit of time, there should be a built-in tool in each language, for example in Dart there is a [Duration class](https://api.dart.dev/stable/1.24.2/dart-core/Duration-class.html).

d-uzlov:
`second` pecision is absolutely enough for any of our needs in the UI part of the app, because we can use `float64`/`double` datatype for it. Well, we already use it in most places.
In fact, it will be even more precise than what we currently use in the app config, for example.
The only thing that needs the best possible precision is the benchmarking itself, which is done by the loadgen and is completely isolated from the UI part of the app.

>To convert between unit of time, there should be a built-in tool in each language, for example in Dart there is a [Duration class](https://api.dart.dev/stable/1.24.2/dart-core/Duration-class.html).

Thank you, I'm much aware of that class and I don't see how it helps in our case.
If anything, `Duration` is not built for precision. Its methods return integer values, which ruins any precision.

Which option is simpler, easier to read and maintain?
1. `1 / latency`
2. `Duration.millisecondsInSecond / latencyMs`
3. `Duration.microsecondsPerSecond / Duration(milliseconds: latencyMs).inMicroseconds`
    * because `Duration.inSeconds` returns `int` value, which is certainly not precise enough, just like you pointed out.
    * `Duration(milliseconds: latencyMs).inMilliseconds` is also much less precise than using plain millis with float datatype.

And also compare:
1. `min_duration: 60`
2. `min_duration_ms: 60000` ← how many zeroes are there? Looks like 5. Or is it 6? Oh, no, it's actually 4...

[...]

Issue #518, opened 2022-09-15
Avoid manually handling of JSON serialization


anhappdev:
We have many classes with `manual` JSON serialization and will have more.
https://github.com/mlcommons/mobile_app_open/search?q=fromJson
https://github.com/mlcommons/mobile_app_open/search?q=toJson

Since we already use [Protocol Buffers](https://developers.google.com/protocol-buffers) in the app (e.g. [mlperf_task.proto](https://github.com/mlcommons/mobile_app_open/blob/master/flutter/cpp/proto/mlperf_task.proto)), maybe we can also switch to use Protocol Buffers. The benefits are described here: [What are the Benefits of Using Protocol Buffers?](https://developers.google.com/protocol-buffers/docs/overview#benefits)

d-uzlov:
Most of the serialization are for exporting the data to firebase via a Node server (== typescript/javascript code).
* I don't see neither typescript nor javascript listed as a an officially supported language
* Firebase internally works with JSON. Even if we somehow make protobuf work with TS/JS, we would get additional conversions `protobuf → json` and `json → protobuf` with possible bugs and unwanted side-effects.

Also, Dart can't work with the text form of protobuf. We use a special native function in the backend bridge because of it:
https://github.com/mlcommons/mobile_app_open/blob/264e8c13587612a34fee3090c5a3ac884c6870bd/flutter/cpp/flutter/dart_mlperf_config.cc#L13-L15
Not sure how often we will really need the text form of data but it's very convenient for logging and debugging.

Maybe it would be convenient to have declarations of data classes in a separate file, without some language specific stuff (or maybe it wouldn't be). But I think it doesn't justify potential compatibility issues with protobuf either way.

d-uzlov:
I do agree that we have uncomfortably large amount of JSON serialization code in the repo.
I think it could solved by using code generation.
Particularly there is this library: https://pub.dev/packages/json_serializable

d-uzlov:
> Since we already use [Protocol Buffers](https://developers.google.com/protocol-buffers) in the app (e.g. [mlperf_task.proto](https://github.com/mlcommons/mobile_app_open/blob/master/flutter/cpp/proto/mlperf_task.proto)

By the way, if anything, I would like to switch from protobuf to YAML for that particular file. Protobuf is not built for configuration files.
I don't see a point spending time on this, though, since it's just one file which we rarely edit.

anhappdev:
[...]

PR #523, opened 2022-09-19
Change default time unit from `milliseconds` to `seconds`

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/489

### DIFF
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -533,8 +533,7 @@ class BenchmarkState extends ChangeNotifier {
           startDatetime: performance.startTime,
           loadgenInfo: BenchmarkLoadgenInfo(
             validity: performanceInfo.loadgenInfo!.validity,
-            durationMs: Duration.millisecondsPerSecond *
-                performanceInfo.loadgenInfo!.meanLatency *
+            duration: performanceInfo.loadgenInfo!.meanLatency *
                 performanceInfo.loadgenInfo!.queryCount,
           ),
         ),
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -372,8 +372,12 @@ class BenchmarkState extends ChangeNotifier {
       final perfTimer = Stopwatch()..start();
       progressInfo.accuracy = false;
       progressInfo.calculateStageProgress = () {
-        final minDuration = max(benchmark.taskConfig.minDurationMs, 1);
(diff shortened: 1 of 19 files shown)


flutter_common/lib/data/results/benchmark_result.dart
```diff
@@ -34,25 +34,25 @@ class Accuracy {
 
 class BenchmarkLoadgenInfo {
   static const String _tagValidity = 'validity';
-  static const String _tagDuration = 'duration_ms';
+  static const String _tagDuration = 'duration';
```
anhappdev, line 37: What happens with old results when we changed the tag here? Is it still correct?
For now, this may not be a big issue but later if the results are collected, we need a migration plan.

flutter/lib/benchmark/state.dart
```diff
@@ -359,7 +359,9 @@ class BenchmarkState extends ChangeNotifier {
         progressInfo.cooldown = true;
         final timer = Stopwatch()..start();
         progressInfo.calculateStageProgress = () {
-          return timer.elapsedMilliseconds / progressInfo.cooldownDurationMs;
+          return timer.elapsedMilliseconds /
```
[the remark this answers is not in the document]
d-uzlov: I didn't use it here because `inSeconds` getter returns int.
But, come to think of it, seconds precision should be enough for UI, so I guess it will be OK to change.

flutter_common/lib/data/results/benchmark_result.dart
d-uzlov: Old results will fail to parse from JSON.

> but later if the results are collected, we need a migration plan

Yeah. We will need to be more careful with the JSON structure and create migration plans after the next release.

anhappdev (approved):

flutter_common/lib/data/results/benchmark_result.dart
anhappdev: If the `tag` does not match, the JSON file should still be parsed, but the unmatched `tag` will be empty. This should be the default behavior instead of dropping the old results and create a new JSON file. I don't know yet how to do that Dart (but did that in other languages) so we can create an issue for this.

anhappdev: https://github.com/mlcommons/mobile_app_open/issues/518#issuecomment-1250741946

anhappdev (approved):

Issue #475, opened 2022-08-15
Allow to build ios version without all backends


d-uzlov:
Currently coreml backend is enabled automatically when building on macos:

https://github.com/mlcommons/mobile_app_open/blob/29cb9deb341709cee764325ea8e568ae0f8df892/mobile_back_apple/apple_backend.mk#L16-L22

We can use a fake empty backend to substitute missing frameworks when tflite or coreml backends are disabled.

PR #524, opened 2022-09-19
Allow to build ios version without all backends

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/475


anhappdev:
I would use `dummy` instead of `fake`, sound a bit better 😄, just a personal choice, though.

d-uzlov:
Yeah, I didn't know how to name it, so I came up with `fake`. But your idea does sound better :-)
Thank you, I'll update the PR.

anhappdev (approved):

PR #527, opened 2022-09-22
Update Flutter package and some minor fixes

anhappdev:
* Replace deprecated package `device_info` with `device_info_plus`. `device_marketing_names` has `device_info_plus` as a dependency anyway, so we have a package less. And less dependency is always good.
* Inherit deployment target of app for each installed pod.


flutter/lib/device_info.dart
```diff
@@ -63,10 +63,10 @@ class DeviceInfo {
     final deviceNames = DeviceMarketingNames();
 
     return DeviceInfo(
-      manufacturer: deviceInfo.manufacturer,
-      modelCode: deviceInfo.model,
+      manufacturer: deviceInfo.manufacturer ?? 'Unknown',
+      modelCode: deviceInfo.model ?? 'Unknown',
```
d-uzlov, line 67: I think it would be better to use an empty string instead of unknown, and then transform it to `unknown` or any other value in UI.
Checking for an empty string would be easier and less error-prone than comparing to `Unknown` if we want to do any processing of missing values in the future.

anhappdev: Empty string is convenient but I try to avoid using it when we have nullable type.
I think we can make the `manufacturer` and `modelCode` nullable, and keep it nullable until we really need to convert it to a string (`""` or `"unknow"`) like showing in the UI or export to JSON?

d-uzlov: Yeah, making it nullable makes sense.

anhappdev: Done

d-uzlov (approved):

PR #530, opened 2022-09-22
Improve code analysis workflow

anhappdev:
* Automatically add a comment with URL to the SonarCloud dashboard on each new PR.
* Make the docker image tag used in Makefile a variable.

### DIFF
--- a/flutter/android/android-docker.mk
+++ b/flutter/android/android-docker.mk
@@ -32,8 +32,6 @@ flutter_common_docker_flags= \
 		--rm \
 		${flutter_docker_tty_arg} \
 		--init \
-		--user `id -u`:`id -g` \
-		--env USER=mlperf \
 		-v $(CURDIR):/mnt/project \
 		--workdir /mnt/project \
 		-v /mnt/project/flutter/build \
--- a/flutter/android/android-docker.mk
+++ b/flutter/android/android-docker.mk
@@ -32,6 +32,8 @@ flutter_common_docker_flags= \
 		--rm \
 		${flutter_docker_tty_arg} \
 		--init \
+		--user `id -u`:`id -g` \
+		--env USER=mlperf \
 		-v $(CURDIR):/mnt/project \
 		--workdir /mnt/project \
 		-v /mnt/project/flutter/build \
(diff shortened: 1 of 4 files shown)


flutter/android/android-docker.mk
```diff
@@ -30,8 +32,6 @@ flutter_common_docker_flags= \
 		--rm \
 		${flutter_docker_tty_arg} \
 		--init \
-		--user `id -u`:`id -g` \
-		--env USER=mlperf \
```
anhappdev, line 34: @d-uzlov 
1. Do you know why we need `--user ...` here? The docker container is sandboxed so using root would be ok and can avoid some permission problem, though.
2. The `--env USER=mlperf` seems to be from legacy code, I don't see it used anywhere.

d-uzlov: We mount a host volume into docker.
If the program in docker creates any files in the mounted volume, those files will be owned by root.
I don't remember exactly how much problems this brings, but at least all build files would probably require sudo to delete or change them.

>The `--env USER=mlperf` seems to be from legacy code, I don't see it used anywhere.

As far am I'm aware, currently the only thing it does is it's removes "unknown user" warning if you open the terminal inside the docker image.

tools/scanner/scan.mk
```diff
@@ -32,7 +33,10 @@ endif
 
 # Use the same image tag used in `flutter_common_docker_flags`
```
d-uzlov, line 34: Why do we use the same image tag here?

flutter/android/android-docker.mk
anhappdev: Ok. I added it back. I tried to cache Bazel output so I need to remove it else the docker container cannot access the file (see https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user). But caching Bazel output makes the scan command fail, so no need to remove it now.

tools/scanner/scan.mk
anhappdev: The image tag is defined in the `flutter_common_docker_flags`. The commands in `scan.mk` uses it to run the container. Since I don't want to change the existing commands so I just build a new image with some extra tools needed for code scanner and use the same tag.

d-uzlov (approved):

PR #531, opened 2022-09-22
Use code generation for json serialization

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/518


anhappdev:
What happen if
1. The JSON file has additional field which is not existed in the wrapper class?
2. The JSON file doesn't have a field which is existed in the wrapper class?

Can you also add a unit test for import and export from a JSON file with 2 cases above?

d-uzlov:
> 1. The JSON file has additional field which is not existed in the wrapper class?

It will be silently ignored.
There is an option for this in the serialization library: [disallowUnrecognizedKeys](https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonSerializable/disallowUnrecognizedKeys.html), but the default behavior is what we need.

> 2\. The JSON file doesn't have a field which is existed in the wrapper class?

An exception will be thrown.
We can avoid it if we make the [possibly] missing field nullable, but then we would have to account for the possibility of null.
Alternatively we can provide a default value, if it is appropriate for some field.

> Can you also add a unit test for import and export from a JSON file with 2 cases above?

What would be the point?
Serialization code is generated by a library, and we can expect that the library will maintain its behavior.

We can use the test as documentation but the issue is that `disallowUnrecognizedKeys` can be set separately for each class, and optional fields (nullable or with default value) are controlled per field.
So it would be very easy to change the actual behavior without affecting the test.

I think unit tests won't really help us here.

---

By the way, we already have a kind of an indication that something changed in the form of `extended-result.schema.json` file. It's not a unit test but any changes in it is a flag to think about possible break of compatibility.

anhappdev:
> There is an option for this in the serialization library: [disallowUnrecognizedKeys](https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonSerializable/disallowUnrecognizedKeys.html), but the default behavior is what we need.

Default behavior can be changed so if we rely on it, we should either make it explicit, or have a test case to cover it.

> 2. The JSON file doesn't have a field which is existed in the wrapper class?
> An exception will be thrown.

That means if we add a new field (which we often do) the app we create a new JSON file and drop the old results since we catch the exception? I don't think it's a behavior we want. The logs are still there, yes, but it's not convenient anymore like viewing the history in the app.

> Can you also add a unit test for import and export from a JSON file with 2 cases above?
> What would be the point?

I would take a JSON file generated from the app, then add or delete some fields to test it behavior. So when we add or delete or modify a field in the code, the test would tell us that it still behaves as expected. Default behavior from library can be changed, we may not always read the docs carefully when updating the version. The test also helps us to know that it is expected behavior. I will not know about it without asking you or reading the documentation for the library. And after several months I will probably forget it again.

d-uzlov:
> That means if we add a new field (which we often do) the app we create a new JSON file and drop the old results since we catch the exception? I don't think it's a behavior we want.

If we want a different behavior we need a migration policy. It will require time and effort.
We will have to do it for release builds.
But I don't think it worth doing it for intermediate developer builds.

It would be especially wasteful right now, because we are making a lot of changes to make filtering work.

Later, just before we make a new public release, we can take the previous release, save result sample from it, create a migration function, and create a unit test of migration.

d-uzlov:
We will also need some versioning for result spec, to understand how to make a migration.
But I think we should create a separate issue for it, to make this PR simpler.

d-uzlov:
We will also need to better handle cases where result is failed to parse. We don't have to drop such results.
We will also need to stop using `.json` file as a database of results.
But I don't think any of it should be a part of this PR.

anhappdev:
Adding new fields is a very common scenario. All DB system will do it without any issue. I don't think we need a migration plan for adding a new field. Making all fields nullable with a default value would work for me.

d-uzlov:
I created an issue to discuss possible migration plans:
* https://github.com/mlcommons/mobile_app_open/issues/534

anhappdev:
> I created an issue to discuss possible migration plans:
> 
> * [Develop migration plan strategy for exported results #534](https://github.com/mlcommons/mobile_app_open/issues/534)

Please add tests for cases:
1. add a new field
2. delete an existing field
3. rename an existing field

We can update the test after discussion how to handle those automatically, but in this PR the test should cover the expected behavior for those cases.

d-uzlov:
@anhappdev, done.

flutter_common/test/data/serialization_test.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:mlperfbench_common/data/results/backend_info.dart';
+
+void main() {
+  group('json serialization', () {
+    final example = BackendReportedInfo(
+      acceleratorName: '1',
+      backendName: '2',
+      filename: '3',
+      vendorName: '4',
+    );
```
anhappdev, line 12: I prefer to test the whole generated `results.json` file. So if anyone refactors the code, like rename a field, the test would fail and we know that a migration is needed.

anhappdev (changes requested):

anhappdev (approved):

PR #559, opened 2022-10-07
Run iOS tests with GitHub Actions

anhappdev:
* Add a GitHub workflow to run iOS test
* Cache Bazel output in XCode Cloud

### DIFF
--- a/flutter/ios/ci_scripts/ci_post_clone.sh
+++ b/flutter/ios/ci_scripts/ci_post_clone.sh
@@ -1,11 +1,12 @@
 #!/bin/sh
 
-# This script runs after the repo is cloned and before any action execution.
+# This script runs after the repo is cloned and before any action execution in Xcode Cloud
 # Environment variables with `CI_` prefix are set by Xcode Cloud
 # (see https://developer.apple.com/documentation/xcode/environment-variable-reference)
+# Environment variables with `GITHUB_` prefix are set by GitHub Actions
+# (see https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables)
 # To avoid collision with other environment variables, we use `MC_` (MLCommons) prefix for our own variables.
 # This script can also be run on the local Mac machine for testing and debugging.
-# We use [ "$CI" = "TRUE" ] to test if the script is run on local machine or Xcode Cloud.
 
 # abort if an error occurred, show cmd executed
 set -ex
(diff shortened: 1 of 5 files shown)


d-uzlov (approved):
I think it would be better to split `ci_post_clone.sh` file into several smaller scripts, each doing one small thing, to make it easier to read and support.
It would also be nice to split `Run test` github step into several steps: at least `Install dependencies`, `Build native code`, `Prepare Flutter`, `Run test`. It would be easier to read logs this way, and easier to understand what failed.

But I don't think it's that important right now, we can refactor this later.
LGTM.

flutter/ios/ci_scripts/ci_post_clone.sh
```diff
@@ -62,18 +103,27 @@ cd "$MC_REPO_HOME"/flutter && flutter precache --ios
...
+  else
+    cd "$MC_REPO_HOME"/flutter && flutter build ios --config-only --dart-define=official-build="$OFFICIAL_BUILD"
+  fi
 fi
 
 cd "$MC_REPO_HOME"/flutter/ios && pod install
 
+if [ $runner = $GITHUB_ACTIONS ]; then
+  open -a Simulator
+  cd "$MC_REPO_HOME" && make flutter/test
```
d-uzlov, line 125: By the way, `make flutter/test` also runs unit tests.
I guess there's no real harm in running them, they are quite fast, but unit test are platform-independent, so there is no difference between running them here and running them in the `Unit tests / flutter` github workflow.

d-uzlov:
> Part of #121

This is quite a bit part. Also not really a part of "investigating", more of an improvement/refinement IMO.
Maybe we should create separate issues for such stuff, to make it easier to track.

anhappdev:
> I think it would be better to split `ci_post_clone.sh` file into several smaller scripts, each doing one small thing, to make it easier to read and support. It would also be nice to split `Run test` github step into several steps, at least these ones: `Install dependencies`, `Build native code`, `Prepare Flutter`, `Run test`. It would be easier to read logs this way, and easier to understand what failed.
> But I don't think it's that important right now, we can refactor this later. LGTM.

Because GitHub Actions does not cache finished steps like in Dockerfile, split them into smaller steps is mostly visual, since if the workflow fails, one need to check the detailed log anyway. But yes, it would be nice to see the workflow in multiple steps, the trade-off is more code to pass the variable between steps, since each step run in it own shell.

> Part of https://github.com/mlcommons/mobile_app_open/issues/121
> This is quite a bit part. Also not really a part of "investigating", more of an improvement/refinement IMO.

Yes, a separated issue would be better.

flutter/ios/ci_scripts/ci_post_clone.sh
anhappdev: Ideal case there should be no difference, but they are run on 2 different OS (Mac and Linux) so let it run on both would be better.

PR #571, opened 2022-10-12
Add PR decoration for SonarScanner

anhappdev:


anhappdev:
Cannot get pull request decoration to work. It seems like that it's available only on a paid plan.

> Pull Request analysis is available starting in [Developer Edition](https://redirect.sonarsource.com/editions/developer.html).
https://docs.sonarqube.org/latest/analysis/pull-request/

.github/workflows/code-analysis.yml
```diff
@@ -1,10 +1,6 @@
 name: Code Analysis
 
-on:
-  push:
-    branches: [ master, android-v2 ]
-  pull_request:
-    types: [ opened, synchronize, reopened ]
+on: [push, pull_request]
```
d-uzlov, line 3: `on push` will trigger this workflow for each commit, where `code-analysis.yml` is present.
If we do want this, we should remove `on pull_request`, since it triggers this workflow second time for each commit in pull requests.

anhappdev: Thanks for the catch. Updated it.

d-uzlov:
> Kudos, SonarCloud Quality Gate passed!

@anhappdev, are we going to get such comments from sonar bot for each commit, or is it just for the first commit in a pull request?

anhappdev:
> > Kudos, SonarCloud Quality Gate passed!
> 
> @anhappdev, are we going to get such comments from sonar bot for each commit, or is it just for the first commit in a pull request?

I don't know yet. Best case it would auto-update the comment on each new commit like the cla-bot. Let test it out with a new commit.

anhappdev:
It deletes the old comment and posts a new one with updated status. So I think it's fine.

d-uzlov (approved):

PR #576, opened 2022-10-16
Rename BenchmarkSetting's field `configuration` to `framework`

### LINKED ISSUE #540: Make field name in BenchmarkSetting less confusing
Currently:
1. The `configuration` field is used to describe the ML framework (the software).
2. The `accelerator` field is sometime used to describe the framework (the software) and sometime the hardware.
3. The `accelerator_desc` field has mostly the same value as in `accelerator`.

This caused confusing in many places, so I propose the following changes:
1. Rename  `configuration` field to `framework` to describe the software used.
2. Update the `accelerator` field to describe only the hardware used.
3. Remove the `accelerator_desc` field since it mostly has the same value as in `accelerator` field.

https://github.com/mlcommons/mobile_app_open/blob/775ab8b240b0cecbec575c332372395975ce27c8/flutter/cpp/proto/backend_setting.proto#L36-L55

anhappdev:
Closes #540


mohitmundhragithub (dismissed):

flutter/integration_test/first_test.dart
```diff
@@ -54,19 +54,12 @@ void checkAccuracy(BenchmarkExportResult benchmarkResult) {
   );
   expectedMap!;
 
-  // backends are not forced to follow backendSettingsInfo values
-  // we should use backendInfo.accelerator, it always represents real accelerator value
-  var accelerator = benchmarkResult.backendInfo.acceleratorName;
-  if (accelerator == 'ACCELERATOR_NAME') {
```
d-uzlov, line 60: Why is this removed?

mobile_back_pixel/cpp/backend_tflite/tflite_settings_pixel.h
```diff
@@ -29,55 +29,55 @@ common_setting {
 
 benchmark_setting {
   benchmark_id: "image_classification"
-  accelerator: "nnapi"
-  accelerator_desc: "NNAPI"
-  configuration: "TFLite"
+  accelerator: "tpu"
```
d-uzlov, line 32: `accelerator` is not a human-readable string that can be changed to match `accelerator_desc`.
It's a string that is used during backend configuration, so changing it changes behavior of backend.

Particularly this change from `nnapi` to `tpu` makes the backend fall back to CPU because `tpu` is not a valid accelerator tag.

See here:
https://github.com/mlcommons/mobile_app_open/blob/375bac4c0400d9312b059ed0d541c45a658bce54/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc#L286

d-uzlov: Same goes for all the other changes of `accelerator` field in this PR.

d-uzlov (changes requested):

mobile_back_pixel/cpp/backend_tflite/tflite_settings_pixel.h
d-uzlov: You could fix it by editing the backend sources.
But I think it would be better to revert the change.

`nnapi` value reflects the fact that backend uses `tflite::StatefulNnApiDelegate::Options()` during configuration.

`tpu` in pixel config or `gpu+dsp+npu` in tflite android config is just a guess about possible implementation, which may be wrong.
We can probably still use such guesses for `accelerator_desc` field, but I don't think it's suitable for the `accelerator` field.

flutter/integration_test/first_test.dart
anhappdev: It would make the code easier to follow. It would be more simple to enforce backend to set `accelerator` field when adding accuracy test for it.

d-uzlov: It makes sense to add a test for it, to make sure that our default config doesn't use accelerator fallback.

d-uzlov: Maybe it would be better to disable the fallback completely.
Backends should detect supported hardware when choosing the config, and then they should follow the settings, and return an error if something doesn't match.

But this is not related to this PR, we should create a separate issue for this.

mobile_back_pixel/cpp/backend_tflite/tflite_settings_pixel.h
d-uzlov: Using `npu` seems OK to me.

anhappdev: Updated

mobile_back_pixel/cpp/backend_tflite/tflite_pixel.cc
```diff
@@ -172,7 +172,7 @@ mlperf_backend_ptr_t mlperf_backend_create(
       if (strcmp(configs->accelerator, "gpu_f16") == 0)
         options.inference_priority1 = TFLITE_GPU_INFERENCE_PRIORITY_MIN_LATENCY;
       delegate = TfLiteGpuDelegateV2Create(&options);
-    } else if (strcmp(configs->accelerator, "nnapi") == 0) {
+    } else if (strcmp(configs->accelerator, "tpu") == 0) {
       auto options = tflite::StatefulNnApiDelegate::Options();
```
d-uzlov, line 176: I think it would be better to keep the `npu` here.
First of all, pixel backend still uses the `StatefulNnApiDelegate` option, same as tflite backend.

But also, Pixel backend is a copy of tflite with few modifications. I believe that it is regularly updated by copying a fresh version of tflite, and reapplying changes. I don't think we should add add unnecessary changes, it would make updating more error-prone.

d-uzlov: We should probably ask the backend maintainers from Google about this change.

mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h
```diff
@@ -31,36 +31,36 @@ common_setting {
 
 benchmark_setting {
   benchmark_id: "image_classification"
-  accelerator: "coreml"
-  accelerator_desc: "CoreML"
-  configuration: "TFLite"
+  accelerator: "ane"
+  accelerator_desc: "ANE"
+  framework: "TFLite CoreML"
```
d-uzlov, line 36: You should also update the `coreml` check in tflite backend to match new settings, or revert it.

mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h
```diff
@@ -31,36 +31,36 @@ common_setting {
...
 }
 
 benchmark_setting {
   benchmark_id: "natural_language_processing"
-  accelerator: "metal"
-  accelerator_desc: "Metal"
-  configuration: "TFLite"
+  accelerator: "gpu"
+  accelerator_desc: "GPU"
+  framework: "TFLite Metal"
```
d-uzlov, line 54: You should also update the `metal` check in tflite backend to match new settings, or revert it.

By the way, `gpu` value is already used, and it's different from `metal`.

mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h
anhappdev: Done

mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h
anhappdev: Done.

mobile_back_pixel/cpp/backend_tflite/tflite_pixel.cc
anhappdev: The accelerator is changed to `tpu` as requested here in this https://github.com/mlcommons/mobile_app_open/pull/579#issuecomment-1289856432

anhappdev: @kcwu Are you ok with the change here?

flutter/integration_test/first_test.dart
anhappdev: I think we can update the `expectedMap` to `accelerator|framework`
```dart
  final accelerator = benchmarkResult.backendSettings.acceleratorCode;
  final framework = benchmarkResult.backendSettings.framework;
  final expectedValue = expectedMap['$accelerator|$framework'];
  final tag =
      '${benchmarkResult.benchmarkId} [accelerator: $accelerator | framework: $framework]';
  expect(
    expectedValue,
    isNotNull,
    reason: 'missing expected accuracy for $tag',
  );
```
In that case, a fallback will fail this test since it cannot find the expected accuracy

d-uzlov: This will work, for our current setup.

There is nothing preventing you from using the same `framework` value for different accelerators.

In this PR the `framework` field does not match its description. I'm kinda OK with this, but if we were to correctly follow the description of this field both `npu` and `cpu` accelerator values would correspond to the same `TFLite` framework description, for example.
I don't think we should rely on such inconsistent behavior.

d-uzlov: Adding the test that accelerator in config and reported accelerator match is simpler and more reliable.

anhappdev: > Adding the test that accelerator in config and reported accelerator match is simpler and more reliable.

Yes, it is better.
Updated.

flutter/integration_test/first_test.dart
```diff
@@ -42,6 +42,7 @@ void checkTasks(ExtendedResult extendedResults) {
     expect(benchmarkResult.performanceRun!.throughput, isNotNull);
 
     checkAccuracy(benchmarkResult);
+    checkAccelerator(benchmarkResult);
```
d-uzlov, line 45: Please, move accelerator check above accuracy check, so we get "accelerator mismatch" error instead of "missing expected accuracy".

anhappdev: The `checkAccelerator()` cannot work reliable since we do not use `enum` but `string` for the accelerator and the reported value from backend is differently across backend. 
For now, it works only with Pixel 5 with TFLite backend. iOS run in simulator. Pixel 6 with TFLite-Pixel backend still does not implement the accelerator reporting. We don't have physical devices for other vendors.
I think we should just remove this test for now. And add it with a proper implementation later if needed.

d-uzlov: It works on tflite android and tflite windows.
It should also work for iOS in simulator because simulator doesn't use device-specific accelerators:
https://github.com/mlcommons/mobile_app_open/blob/a1f15c945b592d62ff5de2d04c7dfcee075c8a1d/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc#L318-L319

Vedor backends that don't support accelerator reporting we get `ACCELERATOR_NAME` value, which we can check.

anhappdev: The iOS tests currently run with `backend_apple`, so it does not work.
TFLite Android on a simulator will fail (on my test computer at least, the task `natural_language_processing` is expected to use `GPU (FP16)` but backend reported `CPU`)
I do not have windows but it uses `CPU`, so it should pass since there is no accelerator expected.
That means it works only in some cases.
Anyway, if this test should be implemented, we should do it in another PR then.

d-uzlov: > The test if runs on simulator, independently of backend, will fail since we compare the desired accelerator with the actual used accelerator, and simulator does not have accelerator.
> That means it works only in some cases.

No, it doesn't mean it "works only in some cases". It works always.
It _passes_ only in some cases. Which is good: it passes only if the app behavior is what we expect. This is what tests are for.

> Anyway, if this test should be implemented, we should do it in another PR then.

OK, seems reasonable.

d-uzlov: I created an issue for it: https://github.com/mlcommons/mobile_app_open/issues/590

d-uzlov (approved):
LGTM

PR #578, opened 2022-10-16
Rename BenchmarkSetting's field configuration to framework for MediaTek

anhappdev:
Part of #540


mobile_back_tflite/cpp/backend_tflite/neuron/tflite_settings_mtk.h
```diff
@@ -41,7 +41,7 @@ benchmark_setting {
   benchmark_id: "image_classification_offline"
   accelerator: "neuron"
```
anhappdev, line 42: The `accelerator` field should describe the hardware. @freedomtan Can you check if this value is correct?

PR #583, opened 2022-10-21
Set new code options for sonar scanner

anhappdev:
Set new code options for sonar scanner per this guide:
https://docs.sonarcloud.io/improving/new-code-definition/

### DIFF
--- a/tools/scanner/scan.mk
+++ b/tools/scanner/scan.mk
@@ -54,6 +54,7 @@ scanner/scan:
 	sonar-scanner \
 		-Dsonar.organization=mlcommons \
 		-Dsonar.projectKey=mlcommons_mobile_app_open \
+		-Dsonar.projectVersion=${FLUTTER_APP_VERSION} \
 		-Dsonar.sources=. \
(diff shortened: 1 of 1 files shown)


anhappdev:
I would choose `Previous version`.

morphine00:
Ok, that should be set now. Also is there a way that you or someone from the mobile group can manage this instead of me? Would saves us both the time, you having to ask me and me being a carrier pigeon :)

What permissions on the repo are necessary so that I could hand off the SonarCloud settings?

anhappdev:
For SonarCloud, I hope there is no more setting to change, since this one should be "set and forget". But you can assign me an admin role on SonarCloud if you want, since SonarCloud is just a code analysis tool, there is nothing to break.
For GitHub, I will email you when I need more permissions.

morphine00:
Ok, I just granted you permissions on SonarCloud. That was particularly easy.

anhappdev:
Thanks.

d-uzlov (approved):

PR #585, opened 2022-10-24
Remove concurrency group in GitHub Actions

anhappdev:
Remove the concurrency group in GitHub Actions since it cancels running jobs when merging multiples PR into master in a short time.

### DIFF
--- a/.github/workflows/android-tests.yml
+++ b/.github/workflows/android-tests.yml
@@ -1,4 +1,4 @@
-name: Unit tests
+name: Android Tests
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


d-uzlov (approved):

.github/workflows/android-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Unit tests
+name: Android Tests
```
d-uzlov, line 1: This name is misleading.
All of our Android tests run in firebase test lab service.
Tests in this action run in a Dart VM. The VM runs in Linux environment. There is nothing Android-related.

d-uzlov (changes requested):

.github/workflows/android-tests.yml
d-uzlov: By the way, I initially intended this action to also hold any other unit tests. For example, for our C++ code.

d-uzlov (approved):

Issue #556, opened 2022-10-03
Refactor `state.dart` file


d-uzlov:
In previous refactoring we skipped the `state.dart` file.
https://github.com/mlcommons/mobile_app_open/blob/b8c37441fb513a5c3425af37560d0064c27ff445/flutter/lib/benchmark/state.dart

Currently this file has several responsibilities and it's quite big, so we should spend some time refactoring it to make other tasks easier and reduce possibility of introducing bugs.

PR #591, opened 2022-10-26
Refactor `state.dart` (part 1)

d-uzlov:
Part of https://github.com/mlcommons/mobile_app_open/issues/556

In this PR I extract several helper classes from `BenchmarkState`: `TaskRunner`, `ResultHelper`, `ValidationHelper`.

Overall the layout of classes and methods is still suboptimal, but I want to merge this now to reduce possible merge issues.
I have already done manual merge/rebase ~~3 times~~ 4 times because literally any changes to `state.dart` in other PRs require manual merge with this one. And almost any changes in the app require a change in the `state.dart`.

### DIFF
--- a/flutter/lib/backend/bridge/run_result.dart
+++ b/flutter/lib/backend/bridge/run_result.dart
@@ -1,6 +1,6 @@
 import 'package:mlperfbench_common/data/results/benchmark_result.dart';
 
-class RunResult {
+class NativeRunResult {
   final Accuracy? accuracy1;
   final Accuracy? accuracy2;
   final int numSamples;
@@ -10,7 +10,7 @@ class RunResult {
   final String acceleratorName;
   final DateTime startTime;
 
-  RunResult({
+  NativeRunResult({
     required this.accuracy1,
     required this.accuracy2,
     required this.numSamples,
@@ -22,5 +22,5 @@ class RunResult {
   });
 
   @override
-  String toString() => 'RunResult(accuracy:$accuracy1)';
+  String toString() => 'RunResult(accuracy:$accuracy1, accuracy2:$accuracy2)';
 }
--- a/flutter/lib/benchmark/run_mode.dart
+++ b/flutter/lib/benchmark/run_mode.dart
@@ -7,35 +7,40 @@ class BenchmarkRunMode {
   static const _perfLogSuffix = 'performance';
   static const _accuracyLogSuffix = 'accuracy';
 
-  final String mode;
-  final String logSuffix;
+  final String loadgenMode;
+  final String readable;
   final pb.OneDatasetConfig Function(pb.TaskConfig taskConfig) chooseDataset;
 
   BenchmarkRunMode._({
-    required this.mode,
-    required this.logSuffix,
+    required this.loadgenMode,
+    required this.readable,
     required this.chooseDataset,
   });
 
   static BenchmarkRunMode performance = BenchmarkRunMode._(
-    mode: _performanceModeString,
-    logSuffix: _perfLogSuffix,
+    loadgenMode: _performanceModeString,
+    readable: _perfLogSuffix,
     chooseDataset: (task) => task.datasets.lite,
   );
   static BenchmarkRunMode accuracy = BenchmarkRunMode._(
-    mode: _accuracyModeString,
-    logSuffix: _accuracyLogSuffix,
+    loadgenMode: _accuracyModeString,
+    readable: _accuracyLogSuffix,
     chooseDataset: (task) => task.datasets.full,
   );
 
   static BenchmarkRunMode performanceTest = BenchmarkRunMode._(
-    mode: _performanceModeString,
-    logSuffix: _perfLogSuffix,
+    loadgenMode: _performanceModeString,
+    readable: _perfLogSuffix,
     chooseDataset: (task) => task.datasets.tiny,
   );
   static BenchmarkRunMode accuracyTest = BenchmarkRunMode._(
-    mode: _accuracyModeString,
-    logSuffix: _accuracyLogSuffix,
+    loadgenMode: _accuracyModeString,
+    readable: _accuracyLogSuffix,
     chooseDataset: (task) => task.datasets.tiny,
   );
+
+  @override
+  String toString() {
+    return readable;
+  }
 }
--- a/flutter/lib/state/task_runner.dart
+++ b/flutter/lib/state/task_runner.dart
@@ -1,0 +1,295 @@
+import 'dart:io';
+import 'dart:math';
+
+import 'package:async/async.dart';
+import 'package:mlperfbench_common/data/extended_result.dart';
+import 'package:mlperfbench_common/data/meta_info.dart';
+import 'package:mlperfbench_common/data/results/benchmark_result.dart';
+import 'package:uuid/uuid.dart';
+import 'package:worker_manager/worker_manager.dart';
+
+import 'package:mlperfbench/app_constants.dart';
+import 'package:mlperfbench/backend/bridge/isolate.dart';
+import 'package:mlperfbench/backend/list.dart';
+import 'package:mlperfbench/backend/loadgen_info.dart';
+import 'package:mlperfbench/benchmark/benchmark.dart';
+import 'package:mlperfbench/benchmark/info.dart';
+import 'package:mlperfbench/benchmark/run_info.dart';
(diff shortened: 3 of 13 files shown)


d-uzlov:
@anhappdev I think it would be better if you leave your comments, but approve this PR, and then I'll fix it in the next PR for #556.

flutter/lib/state/task_runner.dart
```diff
@@ -0,0 +1,348 @@
...
+
+  Future<RunInfo> run() async {
+    await Directory(logDir).create(recursive: true);
+
+    Cancelable<void>? artificialLoadHandle;
+    if (enableArtificialLoad) {
+      print('Apply the artificial CPU load for ${benchmark.taskConfig.id}');
+      const value = 999999999999999.0;
+      artificialLoadHandle =
+          Executor().execute(arg1: value, fun1: _doSomethingCPUIntensive);
```
anhappdev, line 272: Got this error with the artificial load enabled:
```
flutter: image_classification: performance mode: starting...
[VERBOSE-2:dart_vm_initializer.cc(41)] Unhandled Exception: Invalid argument(s): Illegal argument in isolate message: (object is a SuspendState)
#0      _SendPortImpl._sendInternal (dart:isolate-patch/isolate_patch.dart:249:43)
#1      _SendPortImpl.send (dart:isolate-patch/isolate_patch.dart:230:5)
#2      WorkerImpl.work (package:worker_manager/src/worker/worker_io.dart:60:15)
#3      _Executor._schedule (package:worker_manager/src/scheduling/executor.dart:162:26)
#4      _Executor.execute.executing (package:worker_manager/src/scheduling/executor.dart:125:9)
#5      _Executor.execute.<anonymous closure> (package:worker_manager/src/scheduling/executor.dart:142:71)
#6      Cancelable.thenNext.<anonymous closure> (package:worker_manager/src/cancelable/cancelable.dart:77:27)
<asynchronous suspension>
#7      Cancelable.thenNext.<anonymous closure> (package:worker_manager/src/cancelable/cancelable.dart:73:28)
<asynchronous suspension>
```

This one should be fixed in this PR.

flutter/lib/state/task_runner.dart
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    try {
+      return await _invokeNativeRun();
+    } finally {
+      if (artificialLoadHandle != null) {
+        artificialLoadHandle.cancel();
+      }
```
anhappdev, line 280: Just `artificialLoadHandle?.cancel();` will have the same effect.

flutter/lib/backend/bridge/run_result.dart
```diff
@@ -1,6 +1,6 @@
 import 'package:mlperfbench_common/data/results/benchmark_result.dart';
 
-class RunResult {
+class NativeRunResult {
```
anhappdev, line 3: I think adding a brief description for each of the class as a comment will be helpful since the class name some time is not descriptive enough.

flutter/lib/benchmark/run_mode.dart
```diff
@@ -7,35 +7,40 @@ class BenchmarkRunMode {
   static const _perfLogSuffix = 'performance';
   static const _accuracyLogSuffix = 'accuracy';
 
-  final String mode;
-  final String logSuffix;
+  final String loadgenMode;
+  final String readable;
```
anhappdev, line 11: What does `readable` mean here? Do you mean `description`?

flutter/lib/state/task_runner.dart
```diff
@@ -0,0 +1,348 @@
...
+    }
+    if (benchmark.info.isOffline) {
+      return result.numSamples / loadgenInfo.latency90;
+    } else {
+      return 1.0 / loadgenInfo.latency90;
+    }
+  }
+
+  Future<LoadgenInfo?> extractLoadgenInfo() async {
+    const logFileName = 'mlperf_log_detail.txt';
```
anhappdev, line 336: I think we can make all string `const` like this `static const`.

anhappdev (changes requested):

flutter/lib/state/task_runner.dart
d-uzlov: I changed artificial load stopping back to `Executor().dispose()`, it should be fixed now.

flutter/lib/benchmark/run_mode.dart
d-uzlov: Value of this variable should be human-readable.
Previously it was used only in log file name (which is intended to be handled by vendor developers), now I also use as a log prefix.

Value should be short, to consume less screen space, and it should look nice.
I don't think `description` would fit here, because I would expect description to be more than just one word.

flutter/lib/state/task_runner.dart
d-uzlov: `const` keyword in Dart is not like `const` in C++. It's already `static const`, because in Dart all const values use static initialization, and can only have one instance.
`static const varName` in code will result in compilation error, because it's not a valid syntax in Dart for local fields (for class fields it's the other way around).

flutter/lib/state/task_runner.dart
anhappdev: Please test it to make sure it's really fixed.
BTW, the issue is caused by the method `_doSomethingCPUIntensive()`, which is changed to a class instance method. 
> It needs to be either a static function or a top level function to be accessible as a Flutter entry point.

flutter/lib/benchmark/run_mode.dart
anhappdev: Can you add a comment and example value to explain it? I still not understand what is it and for what it's used?

flutter/lib/state/task_runner.dart
anhappdev: Yeah, `static const` mean the same as `class const` in some other languages, so it needs to be defined as a class field. Normally, I would either define this kind of string as a global constant, or a class constant, since it is kind of setting.

flutter/lib/state/task_runner.dart
anhappdev: You can also add this as a comment to the function `_doSomethingCPUIntensive()`:

> This function needs to be either a static function or a top level function to be accessible as a Flutter entry point.

flutter/lib/benchmark/run_mode.dart
d-uzlov: I think it would be better to change the name than adding comments with example usage.
But I can't come up with a good name.

If you also don't have any fitting suggestions, I'll add a comment in next PR.

flutter/lib/state/task_runner.dart
d-uzlov: OK, I'll move it to class level

flutter/lib/benchmark/run_mode.dart
anhappdev: Maybe just `logSuffix`?

flutter/lib/state/task_runner.dart
d-uzlov: > the issue is caused by the method `_doSomethingCPUIntensive()`

Fixed.

It also seems like `artificialLoadHandle.cancel` doesn't work, so I still kept the `dispose` call.
It seems to work to me now, CPU load changes how I expect it to.

flutter/lib/benchmark/run_mode.dart
d-uzlov: So, just like it was before this PR :-)
I renamed the field because I thought `logSuffix` kinda doesn't fit the fact that it is no longer a true suffix, and the content of the variable will look fitting in any text.

OK, I'll revert it.

anhappdev: I only see it used as a log dir suffix in the code, so that name would be good for now.

d-uzlov: > live logs: `image_classification: performance mode: starting`

Not a log dir suffis

d-uzlov:
@anhappdev could you approve this, please?

anhappdev (approved):

flutter/lib/benchmark/run_mode.dart
anhappdev: I only place I see it used is `logDir = '$logParentDir/${benchmark.id}-${runMode.readable}'`

PR #596, opened 2022-10-29
Refactor state.dart (part 2)

d-uzlov:
Closes https://github.com/mlcommons/mobile_app_open/issues/556

This is a huge pull request, which is not ideal.
If I had enough time I would split it into several small PRs which would be much easier to review.
However, I think it's very important to merge it. I believe that this PR contains a significant code quality improvement (which I was thinking about for the last several months).


flutter/lib/state/app_state.dart
```diff
@@ -0,0 +1,233 @@
...
+
+// All methods in AppState should have the following properties:
+//   - Public methods: return type `void`, public methods must not be awaited
+//   - Public methods: exceptions are never thrown
+//   - All methods: notifyListeners() call outside of `state` setter is forbidden
+//   - All methods: if notifyListeners() is needed when `state` field doesn't change,
+//        the class that contains the data that requires UI update should implement ChangeNotifier instead
+//   - All methods: complex logic should be moved to other classes, methods in AppState should only manage `state`
+//
+// Any methods or getters that don't satisfy these requirements should be subject to refactoring.
```
d-uzlov, line 101: @anhappdev, the #556 issue and this whole PR in particular was for these rules.
Other changes may be subject to discussions, but I think these few lines are very important.
Everything else in this PR was changed to make this possible. Even though there are still few little things left in this class that don't obey these rules, now it will be easy to fix it.

freedomtan:
close this for now. We prefer piecemeal changes. Maybe we can cherrypick some small ones from this huge one later.

Issue #490, opened 2022-08-29
Add filters to online results list


d-uzlov:
We need to add a button for search/filtering of online results, and modify firebase functions API to support it.

Child tasks:
1. Add filtering by OS type
    * https://github.com/mlcommons/mobile_app_open/issues/505
2. Add filtering by backend
    * https://github.com/mlcommons/mobile_app_open/issues/506
3. Add filtering by date
    * https://github.com/mlcommons/mobile_app_open/issues/546
    * Only available when sorting by date
4. Add filtering by manufacturer
    * https://github.com/mlcommons/mobile_app_open/issues/542
    * You can only choose one
    * Also need to maintain a list of known manufacturers
5. Add filtering by exact device model
    * https://github.com/mlcommons/mobile_app_open/issues/543
    * You can only choose one
    * Also need to maintain a list of known models
6. Add filtering by SoC/CPU
    * https://github.com/mlcommons/mobile_app_open/issues/547
    * You can only choose one
    * Also need to maintain a list of known models

We may need to cut some of the filters because of the Firebase limitations.

d-uzlov:
### Firebase limitations

* Support for logical operators `OR` and `!=` is very limited
  * `OR` is only possible once per query and is limited by 10 values in one array field
  * `!=` is only possible once per query
  * `OR` and `!=` are mutually exclusive
* You need to create an index for everything
  * At the very least you need a special index for each queried field and for each sorting option
  * We need 8 possible sorting options: `newest first`, `oldest first`, and 6 `fastest <task> first` (now we have 5 tasks but we expect to add a new one for super resolution)
    * For each field that we add we must add 8 indices
  * Total number of indices is limited by 200
  * It's possible to create indices programmatically

### What this means for us

* Filters of type "select X values from Y possible values" are _generally_ forbidden
  * You can still create such filters provided the number of possible values is known beforehand and is small enough
    * I tried doing this for `OS type` filter, where we only have 3 values
* `Manufacturer`, `Model`, `SoC/CPU` filters must be of type "Select one value from Y possible values"
* We may have up to 25 indexed fields
  * If we add more sorting options this number will decrease
    *  Adding a new task adds a sorting option
  * "select 1 from Y" filter adds 1 indexed field
  * "select X from N" filter adds N indexed fields

freedomtan:
@mkankana for your information, the limitation of Firebase

d-uzlov:
> * We may have up to 25 indexed fields

[...]

PR #623, opened 2022-12-09
feat: store result of each benchmark run in a separated JSON file

### LINKED ISSUE #617: Make a backup of results.json instead of overwriting it
Currently, if we made an update to the data structure of the `results.json` file it will be deleted and new created.
The app should make a backup of the `results.json` file instead of just overwriting it.

This should be a quick fix until https://github.com/mlcommons/mobile_app_open/issues/534 is resolved.

anhappdev:
Instead of storing the results of multiple benchmark runs in one single `JSON` file, the app now stores the result of each run in a separated `JSON` file in the `results` directory.

This should resolve #617 where an existing file is overwritten when failed to parse.
This is also needed to simplify the process of uploading the benchmark result to a server, and also retrieving it from the server for listing and filtering purpose (as in https://github.com/mlcommons/mobile_app_open/issues/490).

### DIFF
--- a/flutter/integration_test/utils.dart
+++ b/flutter/integration_test/utils.dart
@@ -43,9 +43,7 @@ Future<ExtendedResult> obtainResult() async {
   final applicationDirectory =
       await resource_manager.ResourceManager.getApplicationDirectory();
 
-  final rm = result_manager.ResultManager(applicationDirectory);
-  await rm.init();
-
+  final rm = await result_manager.ResultManager.create(applicationDirectory);
   return rm.getLastResult();
 }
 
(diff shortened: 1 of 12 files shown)


freedomtan (approved):

freedomtan:
as far as I can remember, in the submission documentation, we kinda specified that there is a `results.json` which should be pulled from the device and included in submission documents.

@freedomtan to check if we should modify submission documentations

freedomtan:
the doc I mentioned, https://github.com/mlcommons/mobile_open/blob/main/rules/submissionRuleV2_0.adoc

It would be better, if we can have a copy of latest results in `result.json`.

anhappdev:
Thanks for the link. I updated the PR to have a copy of the latest result file always named as `result.json`.

freedomtan (approved):

PR #624, opened 2022-12-09
feat: add new FilterInfo class for result filtering purpose

### LINKED ISSUE #569: Move `selected backend` field in results from benchmark data to top level
Currently we have a separate `selected backend` field for each task because supposedly backends can delegate some tasks to other backends.
However, it's invisible for the application. Application selects only one backend binary at the launch.
The only thing application can see is metadata reported by backend: `backend_name`, `vendor_name`. This metadata will still be per task.

We need `selected backend` field at top level for filtering in online results.

anhappdev:
Closes #569 
Part of #490


anhappdev:
This PR is done, but maybe we should merge it after the code freeze for the submission v3.0.

freedomtan (approved):
LGTM

Issue #638, opened 2023-01-13
Allow PSNR accuracy calculation of Super Resolution task for NCHW returned result


Mostelk:
Currently the accuracy code, reads the ground truth image in in NHWC format, assumes the returned result from backend is also NHWC format. However, some backends are designed for NCHW format. Since is first task we have a 3 channel output result, we need to add this capability during accuracy calculation either in front end, or as a callback function to the backend to reorder its output. Currently preprocesssing allows the backend to order the input https://github.com/mlcommons/mobile_app_open/blob/3576a25ba6c3e4afb5133095ad9c5d24500603b0/flutter/cpp/datasets/snu_sr.cc#L117, so backend can convert NHWC to NCHW. We need accuracy calculation to support both NCHW and NHWC formats.

freedomtan:
Good catch. adding 
```C++
backend_->ConvertInputs(output_size_byte, image_width_, image_height_, ground_truth_vector);
```
after reading a ground truth image should be enough.

However, I don't have the environment to verify it. Could any Samsung colleague help check it. And if it works, send a PR, please.

sb-huh:
Currently there is no way to confirm the data layout of ground truth on vendor side.
In case of handling mass input buffers such as offline, it will occur many redundant conversion of data layout on vendor backend. So I recommend to define explicit definition of data layout on front-end layer and backend owner can refer those definition when it try to conversion its layout.

freedomtan:
Because of the TFLite preprocessing code we use right now, by default, decoded/processed input tensors are in NHWC.

freedomtan:
@freedomtan to check if the TFLite preprocessing code for read input tensors (and ground truth) can do the conversion. (Because we do it in each dataset code). And check if there is other ways to do it (seamlessly).

freedomtan:
Generally, there is no option to turn NHWC raw rgb image to NCHW in tflite preprocessing code [1]. However, load raw RGB code is actually quite trivial [2-3], since we are not doing any real preprocessing for ground truth images, open file stream and then read as is enough.

Anyway, I was wrong about `ProcessOutput()` for each dataset. The `ProcessOutput()` is not called part of LoadGen interface. It's our internal interface. We can modify the interface if all of us agree on what to do.
LoadGen only calls `Name(), TotalSampleCount(), etc` [4]. When output processing completes, we call `QuerySamplesComplete()` [5] to tell LoadGen that outputs are ready.

Outputs are processed then forward

[...]

PR #645, opened 2023-01-24
feat: remove groundtruth_path in lite dataset

anhappdev:
Part of https://github.com/mlcommons/mobile_app_open/issues/638

The lite dataset is used in performance mode, which does not need the ground truth for accuracy test.

### DIFF
--- a/flutter/assets/tasks.pbtxt
+++ b/flutter/assets/tasks.pbtxt
@@ -19,7 +19,7 @@ task {
     lite {
       name: "Open images subset for classification"
       input_path: "https://github.com/mlcommons/mobile_models/raw/main/v0_7/datasets/imagenet.zip"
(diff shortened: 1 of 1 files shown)


mohitmundhragithub:
just removing the groundtruth file for all datasets may not work.
Few of the dataset's code expect the groundtruth to be present, without which they will throw and error i believe.

anhappdev:
> just removing the groundtruth file for all datasets may not work. Few of the dataset's code expect the groundtruth to be present, without which they will throw and error i believe.

Do you know which task specific? I did test run all the tasks in both performance and accuracy mode without any issue.

mohitmundhragithub:
I checked again. Found that the error messages in all the dataset implementations are non-fatal. 
It worked for us as well.

mohitmundhragithub (approved):

PR #659, opened 2023-02-27
chore: use same scripts for format and lint in local and CI environment

### LINKED ISSUE #560: Use the same scripts for format and lint in local and CI environment
When creating a new PR I got an error with the workflow `Format & lint / C++`. It looks like the current used `v4.3.1` of the workflow does not work anymore because the runner image version has changed. Updating the workflow to latest version `v4.9.0` resolves the issue in the old version, but it also adds a check for `.proto` file and fails at that step.

For this particular issue, we can reformat the `.proto` files to pass the test. But it raises another issue, that is the scripts used for `format and lint` are not the same in local environment vs. CI environment. We have a docker container in `tools/formatter` for running `format and lint`, but in CI, we use GitHub actions which run other scripts. Most of the time it works fine, but in some cases like this, the `format and lint` action is passed on local machine but not on CI machine.

My suggestion is that we unify this and use only the `tools/formatter` for both local and CI machine.

anhappdev:
Closes #560


freedomtan (approved):
LGTM

anhappdev:
@freedomtan Can you please re-approve this PR? There was a merge conflict I needed to resolve.

freedomtan (approved):

PR #669, opened 2023-03-21
chore: pin buildifier to specific version

anhappdev:
The CI step [Format & lint / bazel](https://github.com/mlcommons/mobile_app_open/actions/runs/4475740219/jobs/7865521450) of [the last master merge](https://github.com/mlcommons/mobile_app_open/commit/1e03c3b69ced9e5e6e03a9cef16ce466793b2975) failed because the new `buildifier` version reordered the lines.

To fix it, we pin `buildifier` to a specific version to avoid possible issues with new versions.
Also, the flag `--no-install-recommends` was added as recommended by SonarScanner.


freedomtan (approved):
LGTM

PR #674, opened 2023-03-30
feat: add delegate choice

### LINKED ISSUE #662: Add delegate options to the TFLite backend
Currently, we check if the NNAPI Delegate works first. If yes, we use it. If no, falling back to CPU (TFLite interpreter). It would be more convenient if uses could check if she/he wants to use
1. NNAPI delegate,
2. GPU delegate,
3. TFLite interpreter (CPU), or
4. any other 3rd-party delegates

anhappdev:
Closes #662 

### A. Update benchmark_setting
1. Add `delegate_selected` (only one) and `delegate_choice` (one or more) to `benchmark_setting`. 
1. `delegate_selected` must be one of `delegate_choice.delegate_name` else an error will be thrown when the app starts. 
1. The 2 settings are optional, the backend vendor can add it if the user should be able to choose a delegate. 
1. Delegate related settings are now defined in `delegate_choice`. It's recommended to migrate to this new structure.
1. The delegate setting is per benchmark because certain benchmark may only use certain delegate, e.g. `natural_language_processing` can only use `GPU` delegate with the TFLite backend.

Here is the definition:

https://github.com/mlcommons/mobile_app_open/blob/da12e0928d24c0af81907a205e9558e74f3aec33/flutter/cpp/proto/backend_setting.proto#L46-L50

Here is an example:
https://github.com/mlcommons/mobile_app_open/blob/da12e0928d24c0af81907a205e9558e74f3aec33/mobile_back_tflite/cpp/backend_tflite/tflite_settings_android.h#L30-L54

### B. Update backend function:
1. The selected delegate will be passed to the backend when calling `mlperf_backend_create()`, so the backend vendor can use it for creating a new delegate.
https://github.com/mlcommons/mobile_app_open/blob/9b177f30268f7d6ab2da5a2e7ad19d43eea10db8/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc#L204-L216

## UI

The choices are from`delegate_choice`, the default selection is from `delegate_selected`.
When no `delegate_choice` existed, the dropdown will be hidden.

The selected delegate is shown in the result detail screen:


freedomtan:
Something non-trivial:
1. GPUs usually have good fp16 and fp32, but  not int8.
2. recent GPU delegate enable quntized model support by dequantizing a quantized model back to a fp one. Still probably not all quantized ops supported

With that, when we choose to use GPU delegate, we may want to use fp32 models instead quantized ones. Thus we might want to choose fp models.

anhappdev:
> With that, when we choose to use GPU delegate, we may want to use fp32 models instead quantized ones. Thus we might want to choose fp models.

The list of delegates is defined by the vendor in the backend_settings, so theoretically they can define any values they want, like `[ ..., "GPU FP16", GPU FP32", ...]`. Or do you mean we should have a predefined list of delegate for all vendors?

freedomtan:
> > With that, when we choose to use GPU delegate, we may want to use fp32 models instead quantized ones. Thus we might want to choose fp models.
> 
> The list of delegates is defined by the vendor in the backend_settings, so theoretically they can define any values they want, like `[ ..., "GPU FP16", GPU FP32", ...]`. Or do you mean we should have a predefined list of delegate for all vendors?

Nope, I mean for TFLite CPU, we might want to use either https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_uint8.tflite or https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_int8.tflite. But for GPU delegate, mostly https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_float.tflite is a better choice. I don't think we current allow this kind of settings.

anhappdev:
> I don't think we current allow this kind of settings.

Do you mean each delegate choice should have an associated model with it? So, based on the selected delegate, the backend can use a certain model.

anhappdev:
Please review if this updated protobuf structure makes sense.
```protobuf

// Config of a benchmark.
//
// Next ID: 12
message BenchmarkSetting {
  // Id of the benchmark. Must match the value in TaskConfig::id.
  required string benchmark_id = 1;
  // Backend custom model settings
  repeated CustomSetting custom_setting = 2;
  // Human-readable name of the underlying ML framework (software)
  required string framework = 7;
  // single_stream_expected_latency_ns is passed to the MLPerfDriver,
  // default to 1M, max 1M
  optional int32 single_stream_expected_latency_ns = 9 [default = 1000000];
  // Setting for each delegate.
  repeated DelegateSetting delegate = 10;

  // URL or local path of the model file
  // Deprecated, use delegate instead
  required string model_path = 3 [deprecated = true];
  // MD5 checksum to validate the model file
  // Deprecated, use delegate instead
  required string model_checksum = 4 [deprecated = true];
  // Name of the accelerator (hardware), which is passed to the backend.
  // Deprecated, use delegate instead
  required string accelerator = 5 [deprecated = true];
  // Human-readable name of the accelerator (hardware)
  // Deprecated, use delegate instead
  required string accelerator_desc = 6 [deprecated = true];
  // batch_size is passed to the backend
  // Deprecated, use delegate instead
  optional int32 batch_size = 8 [deprecated = true];
}

// Config of a delegate.
// Next ID: 3
message DelegateSetting {
  // Name of the delegate (software)
  required string delegate_name = 1;
  // Name of the accelerator (hardware)
  required string accelerator_name = 2;
  // Human-readable name of the accelerator (hardware)
  required string accelerator_desc = 3;
  // URL or local path of the model file
  required string model_path = 4;
  // MD5 checksum to validate the model file
  required string model_checksum = 5;
  // The batch size to be used when running the model.
  optional int32 batch_size = 8;
}
```

mohitmundhragithub:
can we have CustomSetting as part of DelegateSetting as well? 
Sometimes the customSetting varies based on the delegate that gets selected.

// Backend custom model settings
  repeated CustomSetting custom_setting = 2;

Rest all seems ok to me.

anhappdev:
> can we have CustomSetting as part of DelegateSetting as well?

Yes.

I updated the quick guide in the first comment. Please take a look there to know how it works now.

mohitmundhragithub:
This PR is showing up lots or other diffs as well. Possible to keep only relevant changes to review?

anhappdev:
@mohitmundhragithub 
I mean I updated the [first comment in this PR](https://github.com/mlcommons/mobile_app_open/pull/674#issue-1646920066) to reflect the latest change regarding how we can add delegate choice.

I just merged the latest master branch, so the code should now show only changes related to this PR.

freedomtan:
@AhmedTElthakeb Please report it's OK or not. If yes, then we can merge this.

freedomtan:
@anhappdev to add more "complete" (delegate, model) list in next PR.

freedomtan:
@Mostelk will check this PR again to see if we need to put backend/delegate names somewhere.

freedomtan:
10 qps -> 40 mins for classification offline model (when CPU is used). Should we change this? Either reduce minimum images to test only for CPU offline model, or put maximum time limit (so that we can terminate earlier). Let's open another issue for this.

freedomtan (approved):

PR #699, opened 2023-04-26
feat: use .pbtxt instead of .h file for backend settings

### LINKED ISSUE #694: Use .pbtxt instead of .h file for backend setting
Currently, the backend settings are defined as `string` in a header file, which will be parsed later in code.
We should use the `.pbtxt` format to store those settings and get the benefit of using an IDE to maintain it (like auto-format, style check, syntax check, etc.)

Example of a `.h` file used to store backend setting: [tflite_settings_android.h](https://github.com/mlcommons/mobile_app_open/blob/1709aaf73ebecbe00ff03ae063f382d46fd6c764/mobile_back_tflite/cpp/backend_tflite/tflite_settings_android.h)

And we should convert it to a `.pbtxt` file similar to this one: [tasks.pbtxt](https://github.com/mlcommons/mobile_app_open/blob/1709aaf73ebecbe00ff03ae063f382d46fd6c764/flutter/assets/tasks.pbtxt)

anhappdev:
Closes #694

The focus was to have a solution which make it easy for vendors to adapt it.
I added a custom Bazel rule, which takes a `.pbtxt` file and converts it to a `.h` file. So actually in the code the settings is still stored as a string const, but now we manage it in a `.pbtxt` file. With that, the IDE should make the editing of the `.pbtxt` file easier.

### 1. Extract the settings from `.h` files to `.pbtxt` files.

The quick way without changing much of the original source code is create a new file for each `const std::string` used. For example the string content of `const std::string tflite_settings_apple_iphoneX` is now extracted to the file  [tflite_settings_apple_iphoneX.pbtxt](https://github.com/mlcommons/mobile_app_open/blob/694/use-pbtxt-for-backend-settings/mobile_back_tflite/cpp/backend_tflite/backend_settings/tflite_settings_apple_iphoneX.pbtxt) and replaced with a generated const.

https://github.com/mlcommons/mobile_app_open/blob/f2c2508b65879ed2532018b6074ff8d5b485c6f4/mobile_back_tflite/cpp/backend_tflite/tflite_settings_apple.h#L15-L33

### 2. Add a build rule to generate header files from pbtxt

Load the custom rule
https://github.com/mlcommons/mobile_app_open/blob/f2c2508b65879ed2532018b6074ff8d5b485c6f4/mobile_back_tflite/cpp/backend_tflite/BUILD#L20

Use it
https://github.com/mlcommons/mobile_app_open/blob/f2c2508b65879ed2532018b6074ff8d5b485c6f4/mobile_back_tflite/cpp/backend_tflite/BUILD#L27-L36

Add it as a dep in `cc_library`
https://github.com/mlcommons/mobile_app_open/blob/f2c2508b65879ed2532018b6074ff8d5b485c6f4/mobile_back_tflite/cpp/backend_tflite/BUILD#L57-L59

To import the generated headers in code, see above in step 1.


freedomtan:
@mohitmundhragithub and @AhmedTElthakeb to check this

Issue #695, opened 2023-04-17
Use PNG instead of raw images 


freedomtan:
Current, we use jpg and raw rgb images for ADE20K and SNUSR because we didn't have PNG support.
Since we merged https://github.com/mlcommons/mobile_app_open/pull/691, it should be possible to PNG for both.

freedomtan:
ADE20K: original in PNG. Maybe we can try do segmentation PNG first. @freedomtan

freedomtan:
To use `.png` to replace `.jpg` input images: adding `.png` to supported file formats is enough, see  e6379eb
To use `.png` for ground truth, couple lines of changes needed, see 2f815fd

On Pixel 7 Pro, with Pixel backend:
|.jpg|.png|
|---:|----:|
| 0.5889| 0.5897|
|<img src="https://user-images.githubusercontent.com/3395998/233523796-232b4fe8-3777-41de-930a-f3d65e0c1234.png" width="200"  />|<img src="https://user-images.githubusercontent.com/3395998/233523902-364185c7-3452-435a-8835-6f632ea9a6d2.png" width="200"  />|


- Use `.png` for input images: better quality.
- Use `.png` for groundtruth: reduced file size: 
  - `.raw`: 512x512x2000 = 524,288,000 bytes.
  - `.png`:  < 10 MiB

freedomtan:
In the group meeting this week, @Mostelk mentioned if the NCHW vs NHWC issue. I read through the [PNG RFC](https://www.rfc-editor.org/rfc/rfc2083), it seems the PNG doesn't NCHW natively. So, for super resolution case, options I can think out right now:

1. use raw format (".rgb8")
2. add some output processing hook as we did for input tensors

Mostelk:
One solution to try is; adopt PNG for input processing only, and leave rgb8 (NHWC and NCHW options) for accuracy calculation as current case.

mohitmundhragithub:
I tried this change: https://github.com/mlcommons/mobile_app_open/compare/use_png_for_ade20k
with our backend, but got very poor accuracy. 

![image](https://user-images.githubusercontent.com/1549325/234016806-7cac7a33-a29e-4f88-acce-e60a05cfdb8d.png)

Will need to debug this further to find out the actual reason for this difference though.

freedomtan:
> I tried this change: https://github.com/mlcommons/mobile_app_open/compare/use_png_for_ade20k with our backend, but got very poor accuracy.
> 
[...]

PR #707, opened 2023-05-09
Use png for ade20k

freedomtan:
Since we already have .png support, we can enable .png for ADE20k. Benefits of using .png

- input data: with .png we can use lossless resized data instead of lossy .jpeg
- label: with .png we can have small data size, previously, we used raw 512x512x1 data. 

See #695 for related discussion.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -53,8 +53,9 @@ http_archive(
     name = "org_tensorflow",
     patch_args = ["-p1"],
     patches = [
-        # Add patch for adding png in tflite evaluation code
+        # Add patches for adding png in tflite evaluation code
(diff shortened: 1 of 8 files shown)


freedomtan:
@freedomtan to add corresponding description to the mobile_open repo

anhappdev:
The CI tests failed because it still use the old `raw` groundtruth.
I uploaded the new PNG based groundtruth here: https://github.com/mlcommons/mobile_models/pull/59

Please update the link after https://github.com/mlcommons/mobile_models/pull/59 is merged.
https://github.com/mlcommons/mobile_app_open/blob/31647b7c5228b663376a547ea6d7e2a47ddb7e36/flutter/assets/tasks.pbtxt#L99

anhappdev (approved):

PR #709, opened 2023-05-09
feat: show selected delegate in main screen result

anhappdev:
Related to #662 
Show selected delegate in main screen result


freedomtan:
I am confused. "TFLite | GPU | CPU" doesn't look right to me.

anhappdev:
> I am confused. "TFLite | GPU | CPU" doesn't look right to me.

The text has this format:
```
<Backend> | <User selected delegate> | <Backend reported accelerator>
```

So it can happen that a user selects a `GPU delegate` but when the backend runs it cannot use that `GPU delegate` and fallback to the `CPU`, then the backend reports the `CPU` as the accelerator used.

The screenshot is from an Android simulator BTW.

It looks quite confusing, though. Should we not show the selected delegate at all? Then we have this format like before:
```
<Backend> | <Backend reported accelerator>
```

freedomtan:
> > I am confused. "TFLite | GPU | CPU" doesn't look right to me.
> 
> The text has this format:
> 
> ```
> <Backend> | <User selected delegate> | <Backend reported accelerator>
> ```

Oh, then I am fine with that. Dunno how to present it to general users though.

freedomtan:
current ideas:

* @Mostelk  showing "<Backend> | <User selected delegate> | <Backend reported accelerator>" at the bottom of this screen.
* show a dialog box explaining the "<Backend> | <User selected delegate> | <Backend reported accelerator>"  before running

freedomtan:
@TheKanter suggested we need some help from UX/UI guys. Maybe from Scopic

mohitmundhragithub (approved):

anhappdev:
As discussed in the meeting, we will merge this and update the UI when needed in another PR.

Issue #702, opened 2023-05-03
Check if we can get rid of manage all file permission


freedomtan:
Currently, we have to manually enable the "Manage All Files Access Permission". That's not intuitive and could not be submit to Google Play Store.

https://github.com/mlcommons/mobile_app_open/pull/621

anhappdev:
I spent more time to resolve the issue but could not success.

Without the permission `MANAGE_EXTERNAL_STORAGE` the app can only read media files (`*.jpg`) but not other file types like `*.pbtxt` or `*.tfrecord`. Still don't know why.

For user convenience, we can use this package https://pub.dev/packages/permission_handler to request `MANAGE_EXTERNAL_STORAGE` permission and it will show the UI for that. But it still doesn't resolve the issue that we still need the `MANAGE_EXTERNAL_STORAGE` permission.

freedomtan:
@freedomtan to test if this package [https://pub.dev/packages/permission_handle](https://pub.dev/packages/permission_handler) work for devices having all file access permission in different places

anhappdev:
I am moving the discussion from https://github.com/mlcommons/mobile_app_open/issues/727#issuecomment-1608830898 to this issue for better tracking.

> With the `all file permission` we used currently, it cannot be published to the Play Store.
> 
> Can we ignore this `all file permission` for Play Store publishing.
> 
> what if we make all the filenames with file extension like .png or .jpg (.txt -> .png). If this gonna work? @anhappdev to check it.

Renaming the file extension to a media file extension like .jpg or .png as @freedomtan suggested does work.
I renamed `coco_val_full.pbtxt` to `coco_val_full.jpg` and the app can now read the file without the read all file permission.

The question now is if and how we use this work-around. Using a not correct file extension is ... well... "not correct".

freedomtan:
Let's do something like `coco_val_full.pbtxt.pseudo.png` for `coco_val_full.pbtxt` so that we can remove the scary "all file permission".

anhappdev:
Noted that if the local path in `tasks.pbtxt` is a directory. All files in that directory must be renamed to have a media file type extension.
For example, in the SNUSR dataset, if a vendor still uses `.rgb8` files instead of `.png`, all `.rgb8` files should add the `pseudo.png` extension.

freedomtan:
We have to remove this permission and either
1. have all the datasets in /sdcard/Android/app_name/ (not /data/app/, @freedomtan to update this to the correct name), or
[...]

PR #712, opened 2023-05-16
feat: add setting to grant manage all file permission on Android

anhappdev:
Related to #702 

Added a new switch to redirect user to the setting screen for managing all file access.

APK for testing: [2023-05-16_mlperfbench-45948d7-t.apk](https://drive.google.com/file/d/1uPCvKc7QcsgaWUbGDM11_mYQ1MgD_72E/view?usp=sharing)


freedomtan:
Yes, this PR works on various platforms (AOSP, some Android distributions in China, Google Pixel phones, etc). However, it's enable only. Once I switched it, switching it off didn't trigger the setting screen to change it.

anhappdev:
> However, it's enable only. Once I switched it, switching it off didn't trigger the setting screen to change it.

The used package does not have any method to explicit show the permission setting screen or undo the permission. It just checks if the permission is not granted then shows the setting screen. 
Maybe we can use another symbol instead of a switch to avoid this confusion. Perhaps a button to grant the permission. And after the permission is granted, just show check mark indicated that permission is granted?

anhappdev:
I replaced the switch with a checkmark, now it looks like this:

freedomtan (approved):

PR #718, opened 2023-05-25
feat: add max_duration flag

### LINKED ISSUE #701: offline image classification is slow for for CPU only cases
As reported by @mohitmundhragithub: when CPU fallback is used for offline image classification, the speed is around 
10 queries per second, which means  it takes around 40 mins for classification offline model. 

Should we change this? Either reduce minimum images to test only for CPU offline model, or put maximum time limit (so that we can terminate earlier).

anhappdev:
Closes #701


freedomtan:
@freedom to test this on a device.

anhappdev:
For documentation:
If the benchmark runs longer than `max_duration`, then it will be stopped and the result is invalid. Invalid result will be marked red.

freedomtan:
> For documentation:
> 
> If the benchmark runs longer than `max_duration`, then it will be stopped and the result is invalid. Invalid result will be marked red.
> 
> 
> 
> 
> 
> 
> 
> 

hmm, offline is a valid one?

freedomtan:
> For the test in the screen shot I set max duration to 80 seconds.

hmm, did I miss something? 24576/26.41 = 930.56 >> 80

anhappdev:
It looks like the `max_duration` did not work in Offline scenario. I'm still trying to figure out why.

anhappdev:
Testing with the latest release (`v3.0`) gave me the same result. `max_duration_ms` setting does not work in Offline scenario.

freedomtan:
A possible alternative: show a dialog box to check if the user wants to stop or continue.

freedomtan:
Because offline is supposed be a single inference, so the max duration doesn't work by design. We should look for other alternatives.

freedomtan (approved):

freedomtan:
We can add an offline solution later.

PR #733, opened 2023-06-15
ci: pin protoc_plugin to v20.0.1 instead of using latest version

anhappdev:
A new `protoc_plugin` v21.0.0 is just released and breaks our CI because we use the previous version v20.0.1, which was released 12 months ago (see https://pub.dev/packages/protoc_plugin/versions).

This PR pins the `protoc_plugin` version to v20.0.1.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Create ephemeral files
         run: |
           flutter config --no-analytics &&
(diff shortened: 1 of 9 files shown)


freedomtan (approved):
Do we have other plugins that we need to have fixed versions?

anhappdev:
> Do we have other plugins that we need to have fixed versions?

I am not sure. I always tried to use a specific version, but we still have many legacy code. It's better just to fix it when the issue comes up, instead of finding it now.

Issue #670, opened 2023-03-22
Add result upload and download feature


anhappdev:
- [X] The app can upload result files to a cloud storage (#703)
- [X] The app can download result files from a cloud storage (only results from the same user_id) (#752)
- [x] Provide option to convert from an anonymous account to a normal account (#757)

Background:
In the past, we tried to use the Firebase Database for storing results but it has very limited capabilities in filtering (see https://github.com/mlcommons/mobile_app_open/issues/490). Since we don't expect the data (the result) to be modified, and the amount of data will not be big, we will be fine with just storing the result as a `json/text` file, then upload it to a cloud storage, and download other results from the cloud storage and do all the filtering locally on-device. The filtering is already implemented in https://github.com/mlcommons/mobile_app_open/issues/641.

PR #752, opened 2023-07-15
feat: add result download

anhappdev:
Part of #670 

When the app starts, it will query the list of online results in Firebase Storage under the same user ID, then download ones, which do not exist in local storage.

### DIFF
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -156,26 +156,25 @@ class BenchmarkState extends ChangeNotifier {
   }
 
   static Future<BenchmarkState> create(Store store) async {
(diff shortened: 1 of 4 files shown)


anhappdev:
As listed in https://github.com/mlcommons/mobile_app_open/issues/670 the next PR will add the option to convert from an anonymous account to a normal account.

freedomtan:
@freedomtan to test it

anhappdev:
To test this feature, you will need to first upload a result, then delete that result from local storage (using any File Manager), then restart the app and check if the result is shown in the History screen (or see in the log console which results are downloaded).

freedomtan (approved):
LGTM

PR #758, opened 2023-07-29
feat: add delegate choices for mobile_back_apple

RSMNYS:
feat: adjusted apple backend with the new delegate choices


anhappdev (changes requested):
1. Please ensure all CI tests are passed first before requesting a review.
2. `mlperf_backend_accelerator_name()` should return the real accelerator used, not the one in the configuration, which will be different in a fallback case.

mobile_back_apple/cpp/backend_coreml/coreml_settings.pbtxt
```diff
@@ -1,58 +1,167 @@
...
   benchmark_id: "image_classification"
-  accelerator: "ane"
-  accelerator_desc: "Apple Neural Engine"
   framework: "Core ML"
-  model_path: "https://github.com/mlcommons/mobile_models/raw/main/v3_0/CoreML/MobilenetEdgeTPU.mlmodel"
-  model_checksum: "39483b20b878d46144ab4cfe9a3e5600"
+  delegate_choice: {
+    delegate_name: "All"
+    accelerator_name: "all"
+    accelerator_desc: "All compute units"
```
anhappdev, line 10: I would replace "all" with "cpu&gpu&ane" to be more descriptive.

mobile_back_apple/cpp/backend_coreml/coreml_settings.pbtxt
```diff
@@ -1,58 +1,167 @@
...
+    delegate_name: "CPU & GPU"
+    accelerator_name: "cpu&gpu"
+    accelerator_desc: "CPU and GPU"
+    model_path: "https://github.com/mlcommons/mobile_models/raw/main/v3_0/CoreML/MobilenetEdgeTPU.mlmodel"
+    model_checksum: "39483b20b878d46144ab4cfe9a3e5600"
+  }
+   delegate_choice: {
+    delegate_name: "CPU & NE"
+    accelerator_name: "cpu&ne"
+    accelerator_desc: "CPU and Neural Engine"
```
anhappdev, line 24: It's more common to write Apple Neural Engine or ANE. So I would rename "cpu&ne" to "cpu&ane".

mobile_back_apple/cpp/backend_coreml/coreml_util.swift
```diff
@@ -110,12 +170,21 @@ public class CoreMLExecutor: NSObject {
...
+  ) throws {
     pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0)
     let config = MLModelConfiguration()
-    config.computeUnits = MLComputeUnits.all
+
+    let unitName = String(validatingUTF8: acceleratorName)
+
+    if let name = unitName, let computationUnit = ComputationUnitNameEnum(rawValue: name) {
+      accelerator = MLCommonsComputationUnit(unitName: computationUnit)
+      config.computeUnits = accelerator.computeUnits
```
anhappdev, line 186: What happens if `.cpuAndNeuralEngine` is selected / configured, but the device doesn't have an ANE?

RSMNYS: The option: .cpuAndNeuralEngine is first available from the IOS 16. So with this all devices have the ane. And the operating system decides on which chip runs the calculation. I think even if it not available if will run on cpu, the same way as .all option is working.

anhappdev: iPhone 8 and some older models can still run on iOS 16 [1] and don't have ANE.
And `mlperf_backend_accelerator_name()` should report the actually used accelerator.

[1] https://support.apple.com/en-gb/guide/iphone/iphe3fa5df43/ios

freedomtan: @freedom to check it on iPhone 8+ to see if it crashed or not.

freedomtan: Confirm that when `cpu and ane` was chosen on iPhone 8+ (a device with ANE but not exposed to developers), CPU was used to run benchmark items. NO, it didn't crash.

anhappdev (approved):

PR #769, opened 2023-08-09
feat: check ANE availability on Apple devices

### LINKED ISSUE #768: Apple backend should check if ANE available
Currently, if an Apple device has no Apple Neural Engine, it falls back to use CPU but still reports ANE as accelerator used.
The Apple backend should check if an ANE is available and report correctly the accelerator used.

anhappdev:
Closes #768

### DIFF
--- a/mobile_back_apple/cpp/backend_coreml/coreml_util.swift
+++ b/mobile_back_apple/cpp/backend_coreml/coreml_util.swift
@@ -116,7 +116,7 @@ private class MLCommonsComputationUnit {
   {
     switch computationUnitName {
     case .all:
-      if Device.hasANE == true {
+      if let hasANE = Device.hasANE, hasANE {
         return MLComputeUnits.all
       } else {
         return MLComputeUnits.cpuAndGPU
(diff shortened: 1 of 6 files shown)


mobile_back_apple/cpp/backend_coreml/coreml_util.swift
```diff
@@ -117,18 +116,21 @@ private class MLCommonsComputationUnit {
...
+      } else {
+        return MLComputeUnits.cpuAndGPU
+      }
     case .cpuAndGPU:
       return MLComputeUnits.cpuAndGPU
     case .cpuAndNeuralEngine:
-      if #available(iOS 16.0, *) {
+      if let hasANE = Device.hasANE, hasANE, #available(iOS 16.0, macOS 13.0, *) {
         return MLComputeUnits.cpuAndNeuralEngine
+      } else {
```
RSMNYS, line 129: We can refactor here slightly, and remove the else. So 
if (expression) { 
    return value
}
return anotherValue

anhappdev: I prefer to have the else here to quickly scan the code and see that there are 2 possible exists.

freedomtan (approved):
LGTM

PR #797, opened 2023-10-03
Update Samsung's private libs in CI build

anhappdev:


anhappdev:
@AhmedTElthakeb Please review and merge if it’s ok.

Issue #806, opened 2023-10-31
To update the UI of app


freedomtan:
1. Let's get figma 

- [ ] design from Hanjo @RSMNYS: please help.

2. Let's check feasibility of implementing Hanjo's design: e.g., is there any clear obstacles. 

- [ ] give feedbacks if something is hard to implement
- [ ] and try to finalize the design
- [ ] scope out the work

3.
4. ....

RSMNYS:
Hi guys! Here is the figma file: https://www.figma.com/file/og9GrHGzMQ9PSp8CFCPVdb/Design-Audit?type=design&node-id=69%3A535&mode=design&t=rWeSndRRQQXSMCK8-1

anhappdev:
Please find the effort estimation here:
https://docs.google.com/spreadsheets/d/13hmH3EpnSo7MLOzukxfqiMmJ37GhRXFZUbBQRMPAhuE/edit?usp=sharing

anhappdev:
- [ ] S1 - Resource loading screen
- [ ] S2 - Start screen
- [ ] S3 - Run progress screen
- [ ] S4 - Result screen
- [ ] S5 - History screen
- [ ] S6 - Settings screen
- [ ] S7 - Menu screen https://github.com/mlcommons/mobile_app_open/pull/815

banjo247:
@freedomtan hi there! I still need to handle the sharing/reporting side of the application. The previous files sent through was either a JSON file which I think was a bit confusing as it covers all parameters even if not used, and then David T sent through some excel file that wasn't entirely clear. Is there some kind of output file I can use to start creating some ideas about reports and report sharing of results?

anhappdev:
@banjo247 The app will upload and share the result via a JSON file. Here is an example: [results.json](https://github.com/mlcommons/mobile_app_open/files/13496265/results.json)
We also have a description of the file here: https://github.com/mlcommons/mobile_app_open/blob/master/docs/result-spec.md

freedomtan:
@banjo247: we used brown and blue to represent unofficial and official releases. Do you have any suggestion for colors other than current brown for "unofficial" builds.

anhappdev:
@banjo247 @RSMNYS  We need
- [ ] a design / color for unofficial build (see https://github.com/mlcommons/mobile_app_open/pull/845)
- [ ] a new design for the history screen (see https://github.com/mlcommons/mobile_app_open/pull/832)

anhappdev:
[...]

PR #832, opened 2024-01-08
ui: new history screen

anhappdev:
* Part of #806
* New design for history screen

The original design has some problems:
- It probably based on an outdated version of the app, where we show a simple list of past run (each run contains multiple benchmarks)
- Since we want to sort and filter the results, we must show it at one deeper level (benchmark instead of run).
- For now, I just updated the old result screen with this new simple design. We can improve it further if our designer has some new design.

 → 

history


anhappdev:
@banjo247 Please check my first comment related to the design.

freedomtan (approved):

freedomtan:
@banjo247 ping, we need your comments

freedomtan:
@anhappdev Let's remove the sorting feature, just sort the dates (either ascending or descending)

anhappdev:
- New history screen https://github.com/mlcommons/mobile_app_open/pull/866

PR #843, opened 2024-01-23
test: add workflow for new test device

anhappdev:


anhappdev:
- Closes in favor of https://github.com/mlcommons/mobile_app_open/pull/842

PR #857, opened 2024-02-20
fix: performance drop caused by UI animation

### LINKED ISSUE #856: Check for performance impact of new UI
@mohitmundhragithub reported a drop in performance and suggest it may caused by the new `InfiniteProgressCircle`.
@anhappdev To send a PR with and without this `InfiniteProgressCircle` for testing.

anhappdev:
- Closes #856 

The progress screen now looks like this and has no animation:

--- 

2024-02-22:
- I added a new setting so we can quickly test different animation styles. For comparison, I added back the old dotted circle animation we had before the UI change.
- In my testing with an old Samsung phone, dotted or gradient circle produce similar throughput, and no animation produces a slightly higher throughput. So we should decide whether we want a progress animation or not, then I will update the code and remove the setting.


mohitmundhragithub:
Thanks a lot. We experimented with these settings. 
1) "No animation" gives the best performance.
2) "dotted circle" gives performance as good as the previous UI.
3) "gradient circle" gives performance quite a bit less than previous UI.

mohitmundhragithub:
> @mohitmundhragithub Thanks for the feedback. BTW, I forgot to mention, if you don't use the release APKs from GitHub but build the app locally, please make sure you run it with the [--profile flag](https://docs.flutter.dev/perf/ui-performance#run-in-profile-mode) `flutter run --profile` when testing performance.

which portion will it enable?

freedomtan:
no animation, but when the go button pressed, make the outer circle bright.

mohitmundhragithub (approved):

mohitmundhragithub:
This PR may be required to be cherry-picked to submission-v4.0 branch as well.

PR #859, opened 2024-02-27
chore: increase Android minSdkVersion from 21 to 30

### LINKED ISSUE #858: Crash on older devices with API level <= 29
While testing the app on the following devices:
- Galaxy S20, API Level 29
- Pixel 3, API Level 28
It crashed with this error:
```
02-20 20:32:22.310: I/flutter(16177): Exception: Invalid argument(s): Failed to load dynamic library 'libbackendbridge.so': dlopen failed: cannot locate symbol "pthread_cond_clockwait" referenced by "/data/app/org.mlcommons.android.mlperfbench-jNF9_cYBtBsy20J-3YxASw==/lib/arm64/libbackendbridge.so"...
```

A quick Google search showing this:
> That API (pthread_cond_clockwait ) is [only available on API 30+ devices](https://cs.android.com/android/platform/superproject/+/master:bionic/libc/include/pthread.h;l=124-125;drc=fa658eb094d915a9640a4d0642adee7667f439e7)

https://stackoverflow.com/a/67796165

We either need to fix this error or increase the min supported API level to 30+.

anhappdev:
* Closes #858

### DIFF
--- a/flutter/android/app/build.gradle
+++ b/flutter/android/app/build.gradle
@@ -42,7 +42,7 @@ android {
 
     defaultConfig {
         applicationId "org.mlcommons.android.mlperfbench"
-        minSdkVersion 21
+        minSdkVersion 30
         targetSdkVersion 34
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName


anhappdev:
@mohitmundhragithub Can you please help fix the crash on QTI backend:
```
02-26 23:36:30.204: F/native(26913): cpp/backend_qti/qti_backend_helper.cc:123 runtime 0 is not available on this platform
```

mohitmundhragithub:
> @mohitmundhragithub Can you please help fix the crash on QTI backend:
> 
> ```
> 02-26 23:36:30.204: F/native(26913): cpp/backend_qti/qti_backend_helper.cc:123 runtime 0 is not available on this platform
> ```

I seem to not have permission to view the log files. Can you please help share it?

anhappdev:
> I seem to not have permission to view the log files. Can you please help share it?

Yes. 
[log.txt](https://github.com/mlcommons/mobile_app_open/files/14432024/log.txt)

mohitmundhragithub:
I see this error is causing the test case to fail:

02-26 23:36:30.139: E/audit(1298): type=1400 audit(1709019390.134:12783): avc:  denied  { read } for  pid=26913 comm="DartWorker" name="adsprpc-smd-secure" dev="tmpfs" ino=1396 scontext=u:r:untrusted_app:s0:c78,c257,c512,c768 tcontext=u:object_r:vendor_xdsp_device:s0 tclass=chr_file permissive=0 SEPF_SM-S908U1_12_0001 audit_filtered

MLPerf is considered an untrusted app without some play store key. Seems like with minSDKVersion set to 30, it doesn't allow to read few of the libraries on the devices, for untrusted app.

mohitmundhragithub:
> I see this error is causing the test case to fail:
> 
> 02-26 23:36:30.139: E/audit(1298): type=1400 audit(1709019390.134:12783): avc: denied { read } for pid=26913 comm="DartWorker" name="adsprpc-smd-secure" dev="tmpfs" ino=1396 scontext=u:r:untrusted_app:s0:c78,c257,c512,c768 tcontext=u:object_r:vendor_xdsp_device:s0 tclass=chr_file permissive=0 SEPF_SM-S908U1_12_0001 audit_filtered
> 
> MLPerf is considered an untrusted app without some play store key. Seems like with minSDKVersion set to 30, it doesn't allow to read few of the libraries on the devices, for untrusted app.
> 
> Can you please share the device's android version? I can check internally to find a workaround for this. Also, is the device on the latest update already?

this could be a plausible reason for it: https://developer.android.com/about/versions/10/behavior-changes-10#execute-permission

anhappdev:
The test device was  a `Galaxy S22 Ultra, API Level 33`

Screenshot 2024-02-29 at 10 46 01

freedomtan:
packing .so file in some other ways? suggestion by @mohitmundhragithub

freedomtan:
@RSMNYS to check if we can find any other way to pack .so file

mohitmundhragithub:
> @RSMNYS to check if we can find any other way to pack .so file

@freedomtan I checked the GeekbenchAI app internally. The way they are packing the libs is exactly similar to how we are packing it in the mlperf app.

freedomtan (approved):

PR #860, opened 2024-02-27
ci: use new device for test-android-apk-tflite

anhappdev:
Replace the Pixel 5 with Pixel 6 for `test-android-apk-tflite`


freedomtan (approved):
LGTM

PR #861, opened 2024-02-29
ci: replace obsolete Windows image

anhappdev:
The old Windows image is not available anymore on Google Cloud and therefore be replaced with another one.


freedomtan (approved):
LGTM

PR #878, opened 2024-05-06
feat: add Firebase App Check

### LINKED ISSUE #844: Set up Firebase App Check
> App Check helps protect your API resources from abuse by preventing unauthorized clients from accessing your backend resources. 

https://firebase.google.com/docs/app-check

This means that a user can only upload benchmark result when using the official app released through Google Play Store or Apple App Store.

anhappdev:
* Closes #844


freedomtan (approved):
LGTM

PR #882, opened 2024-05-18
ci: fix Deploy Play Store workflow

anhappdev:


freedomtan (approved):
LGTM. @freedomtan to check the accuracy of Mobilenet V4 on iPhone ANE again.

PR #883, opened 2024-05-18
chore: migrate Flutter SDK from v3.7 to v3.19

anhappdev:
If you already has an existing dev environment, you may need to do this:
- Download Flutter SDK `v3.19.6` from https://docs.flutter.dev/release/archive and update your `FLUTTER_HOME` env to the new version or update your IDE with the new SDK.
- Update your JDK to `v17` or newer
- Run `dart pub global activate protoc_plugin 21.1.2` to update protoc_plugin version
- Run `docker volume rm mlperf-formatter-pubcache mlperf-formatter-output` to clear the cache for `make docker/format`
- Run `cd flutter/ios && rm Podfile.lock && pod install --repo-update` if you develop for iOS


freedomtan (approved):
GMS (Google Mobile Services?) is not enabled by default on phones purchased in China. Some of them can enable GMS easily; some are really complicated. Does our APP work with a phone without GMS enabled.

PR #895, opened 2024-07-02
feat: use MobileNetV4 Core ML model with NCHW input

anhappdev:
The models can be downloaded from here:
https://drive.google.com/drive/folders/1-Mloub0e41mkYs09i9tQZl95dTnGdZ6f?usp=sharing

### DIFF
--- a/mobile_back_apple/.gitignore
+++ b/mobile_back_apple/.gitignore
@@ -1,4 +1,5 @@
 dev-resources
+models/*.mlpackage
 
 # Created by https://www.toptal.com/developers/gitignore/api/xcode
 # Edit at https://www.toptal.com/developers/gitignore?templates=xcode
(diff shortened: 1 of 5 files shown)


anhappdev:
There are 2 issues:
1. I cannot get a plausible accuracy. I am not sure what is wrong with the code?
2. There is no official way to know which format the input data has or the benchmark_id to know when to trigger `mlperf_backend_convert_inputs()`?

freedomtan:
@freedomtan to check the NHWC -> NCHW conversion.

freedomtan:
@anhappdev this NHWC -> NCHW looks right to me. Maybe you can try to compare with the result of using a single transpose op TFLite model (Yes, it's posssible. E.g., in TFLite label_image, I constructed a TFLite model with only `resize` op to resize the input image to the size I needed, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/examples/label_image/bitmap_helpers_impl.h#L30-L100)

anhappdev:
There was an issue with wrong data type, but I got it working now.

freedomtan (approved):

Issue #902, opened 2024-08-13
Pass multiple model paths to backend


anhappdev:
For the Stable Diffusion benchmark, the app needs to download multiple models and pass them to the backend.

anhappdev:
Until now, we have a workaround for this, that we can zip multiple models into 1 `*.zip` archive, but SD models can be very big (3-4 GB), and download 1 huge file is not a good solution.

anhappdev:
My implementation will not change the API so the function signature will stay the same
```
mlperf_backend_ptr_t mlperf_backend_create(
    const char *model_path, mlperf_backend_configuration_t *configs,
    const char *native_lib_path)
```
The `model_path` received from main.cc or the app can be a file path or a directory path. Assuming for Stable Diffusion task, the `model_path` point to a directory where all model files are located and the file names are known beforehand.
The app will download multiple files and put them into one folder, then pass the folder path as the `model_path` to  the backend.

PR #907, opened 2024-08-29
feat: add support for multiple model files

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/902
* The function signature will stay the same:
```
mlperf_backend_ptr_t mlperf_backend_create(
    const char *model_path, 
    mlperf_backend_configuration_t *configs,
    const char *native_lib_path
)
```
* Here is an example how to have multiple models:
```pbtxt
  delegate_choice: {
    delegate_name: "NNAPI"
    accelerator_name: "npu"
    accelerator_desc: "NPU"
    model_file: {
      model_path: "https://file1.tflite"
      model_checksum: "abc"
    }
    model_file: {
      model_path: "https://file2.tflite"
      model_checksum: "def"
    }
  }
```
  * If there is only one model file, the `model_path` will point to the model file like before.
  * If there are multiple model files, the `model_path` will point to a single directory where all files are located in there.

### DIFF
--- a/flutter/cpp/proto/backend_setting.proto
+++ b/flutter/cpp/proto/backend_setting.proto
@@ -51,7 +51,7 @@ message BenchmarkSetting {
 }
 
 // Config of a delegate.
-// Next ID: 8
+// Next ID: 10
 message DelegateSetting {
   // Priority of the delegate. Used for sorting delegate choices in frontend.
   optional int32 priority = 1 [default = 0];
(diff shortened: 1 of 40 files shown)


freedomtan (approved):

anhappdev:
@freedomtan @mohitmundhragithub @AhmedTElthakeb Please take a look at the first comment to know how to define multiple models in the backend settings.

PR #908, opened 2024-09-02
Suggestions for few changes in coco_gen dataset

mohitmundhragithub:
These are few suggestions for the app for coco-gen dataset

1. for performance runs, no need to run "processOutput" completely. This could be determined by not passing the model file.
2. input to the clip_model should be in the range -1.0 to +1.0

### DIFF
--- a/flutter/assets/tasks.pbtxt
+++ b/flutter/assets/tasks.pbtxt
@@ -303,18 +303,18 @@ task {
     type: COCOGEN
     full {
       name: "COCO validation set for Stable Diffusion"
-      input_path: "https://github.com/anhappdev/tmp/releases/download/5/coco_gen_full.tfrecord"
+      input_path: "https://github.com/anhappdev/tmp/releases/download/5/coco_gen_test.tfrecord"
       groundtruth_path: "local:///mlperf_models/stable-diffusion/clip_model_512x512.tflite"
     }
     lite {
       name: "COCO validation set for Stable Diffusion"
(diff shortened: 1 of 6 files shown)


flutter/assets/tasks.pbtxt
```diff
@@ -303,18 +303,18 @@ task {
     type: COCOGEN
     full {
       name: "COCO validation set for Stable Diffusion"
-      input_path: "https://github.com/anhappdev/tmp/releases/download/5/coco_gen_full.tfrecord"
+      input_path: "https://github.com/anhappdev/tmp/releases/download/5/coco_gen_test.tfrecord"
```
anhappdev, line 306: Should it be?
* full -> coco_gen_full (5000 prompts)
* lite -> coco_gen_test (150 prompts)

mohitmundhragithub: I think that's what was decided right? We will use some random samples from full set for performance test.
But use the 150 prompts for measuring accuracy.
But we can double check in the meeting today.

mohitmundhragithub:
@anhappdev @freedomtan i have fixed the path. Please merge this change.

flutter/assets/tasks.pbtxt
anhappdev: Yes. You are correct. I forget that we use the lite one for performance test.

anhappdev (approved):

PR #930, opened 2024-11-01
fix: resolve crash due to permission denied on Android Play Store version

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/893

1. Set `android:extractNativeLibs="true"` as in https://stackoverflow.com/a/64792194
2. Set `android.bundle.enableUncompressedNativeLibs=false` as in https://stackoverflow.com/a/56551499

### DIFF
--- a/flutter/android/app/src/main/AndroidManifest.xml
+++ b/flutter/android/app/src/main/AndroidManifest.xml
@@ -11,7 +11,8 @@
                    android:maxSdkVersion="29"/>
 
   <application android:label="MLPerf Mobile"
-               android:icon="@mipmap/ic_launcher">
+               android:icon="@mipmap/ic_launcher"
+               android:extractNativeLibs="true">
 
     <!-- TFLite -->
     <uses-library android:name="libOpenCL.so"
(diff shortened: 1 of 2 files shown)


anhappdev:
@mohitmundhragithub Can you please test the Play Store version 4.1.0 (252)?

mohitmundhragithub:
> @mohitmundhragithub Can you please test the Play Store version 4.1.0 (252)?

I tried this, and it seems to work on Samsung Galaxy S23 ultra. 
Seems like this is the right fix.

I had tried 1 as you pointed out in description... but never tried 2. May be both are required together.

anhappdev:
> I tried this, and it seems to work on Samsung Galaxy S23 ultra. Seems like this is the right fix.

Great to hear that. So it works for Qualcomm now.
We need to test on a device with Samsung backend to confirm the fix. @AhmedTElthakeb Can you please help test the fix here.

freedomtan:
Let's wait test from @Mostelk

freedomtan:
@AhmedTElthakeb to test it or ask @Mostelk to test it.

Issue #932, opened 2024-11-12
Create a new "general users friendly" mode


freedomtan:
- basic/easy/layman/quick/fast/lite-bench/rapid mode? 
- things to change
  - cool-down / waiting time shorten to 1 min,
  - reduce number of samples (and the minimum running time for each benchmark item)
  - total time to be 5 ~ 10 mins (same constraint for both premium and mid-end devices)
  - for this mode, maybe statistic meaningful is not that important? or we can use mean or median instead of 90 percentile.

Mostelk:
Per discussion with Scott,  Quick Bench 
Target for the whole run to be less than 10 minutes
It is OK to target for Premium Tier first

freedomtan:
- make it possible to run on oldest Pixel phones (Pixel 6) for > 64 iterations, 128 iterations.
- to change parameters, we can replace the https://github.com/mlcommons/mobile_app_open/blob/master/flutter/assets/tasks.pbtxt

Mostelk:
@freedomtan @quic-pnoorzad @mohitmundhragithub @Mostelk come up with some suggestion
~ min time 
~wait time
~min samples
https://github.com/mlcommons/mobile_app_open/blob/master/flutter/assets/tasks.pbtxt 

@anhappdev should we start making a parallel Fast Bench Mode for this that does not overwrite the main Performance and Submission modes

anhappdev:
> @anhappdev should we start making a parallel Fast Bench Mode for this that does not overwrite the main Performance and Submission modes

Ok. I will start working on it.

mohitmundhragithub:
> @freedomtan @quic-pnoorzad @mohitmundhragithub @Mostelk come up with some suggestion ~ min time ~wait time ~min samples https://github.com/mlcommons/mobile_app_open/blob/master/flutter/assets/tasks.pbtxt
> 
> @anhappdev should we start making a parallel Fast Bench Mode for this that does not overwrite the main Performance and Submission modes

I think we can have 2 different tasks.pbtxt files. 1 for the quick bench and another for submission / performance. Would be easier to maintain the settings that way.

- If we want to have a quick benchmark, then probably we don't want to stress the system?
- So both, minimum time and minimum samples could be reduced I think. May be 20secs of min time and 100 minimum samples?
- With less stress on the system, wait time could also be reduced from current 5 mins to may be 1-2 mins? But may need more experimentation.
[...]

PR #941, opened 2024-12-28
feat: add quick run mode

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/932

- The "Integration Test Run" mode will be hidden from user. I just show it here for debug purpose.
- The cool down duration slider will be updated when the user change the run mode. But the duration can still be manually changed by the user.
- The other settings (min_query_count, min_duration, max_duration) are defined in the `tasks.pbtxt` file and cannot be changed by the user.

### DIFF
--- a/flutter/assets/tasks.pbtxt
+++ b/flutter/assets/tasks.pbtxt
@@ -20,8 +20,8 @@ task {
     }
     rapid {
       min_query_count: 64
-      min_duration: 1
-      max_duration: 600
+      min_duration: 6
+      max_duration: 60
     }
   }
   datasets {
@@ -77,8 +77,8 @@ task {
     }
     rapid {
       min_query_count: 64
-      min_duration: 1
-      max_duration: 600
+      min_duration: 6
+      max_duration: 60
     }
   }
   datasets {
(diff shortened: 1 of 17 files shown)


flutter/assets/tasks.pbtxt
```diff
@@ -4,12 +4,26 @@
...
+      min_query_count: 128
+      min_duration: 6
+      max_duration: 60
+    }
+    rapid {
+      min_query_count: 64
+      min_duration: 6
+      max_duration: 60
+    }
+  }
```
anhappdev, line 26: We can define the settings for each run mode in the `tasks.pbtxt` file.

freedomtan:
@AhmedTElthakeb please help test this.

flutter/assets/tasks.pbtxt
freedomtan: please change the `min_duration: 6` to `min_duration: 10` for the quick mode

freedomtan:
Let's try show which mode the app is running (so that we can tell it from the screenshot).

flutter/assets/tasks.pbtxt
freedomtan: for stable diffusion: max duration 40 seconds.

anhappdev: The changes will be in https://github.com/mlcommons/mobile_app_open/pull/947

PR #954, opened 2025-01-30
Add offline dialog

### LINKED ISSUE #948: App should notify users that the device is offline when trying to download files
Currently the device log will show an exception, and the download will stop, but nothing happens in the UI.

farook-edev:
Added Dialog for when an offline (not connected to the internet) user attempts to download resources, the error only occurs in case a SocketException is caught, any other exceptions are still unhandled.

This fixes #948

image

### DIFF
--- a/flutter/lib/benchmark/state.dart
+++ b/flutter/lib/benchmark/state.dart
@@ -143,15 +143,21 @@ class BenchmarkState extends ChangeNotifier {
       modes: [taskRunner.perfMode, taskRunner.accuracyMode],
       benchmarks: benchmarks,
     );
-    await resourceManager.handleResources(
-      resources,
-      needToPurgeCache,
-      downloadMissing,
-    );
-    print('Finished loading resources with downloadMissing=$downloadMissing');
-    error = null;
-    stackTrace = null;
-    taskConfigFailedToLoad = false;
+    try {
+      await resourceManager.handleResources(
+        resources,
+        needToPurgeCache,
+        downloadMissing,
+      );
+      print('Finished loading resources with downloadMissing=$downloadMissing');
+      error = null;
+      stackTrace = null;
+      taskConfigFailedToLoad = false;
+    } on SocketException catch (e, s) {
+      print('Could not load resources due to error: $e');
+      error = e;
+      stackTrace = s;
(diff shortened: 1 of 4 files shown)


flutter/lib/ui/settings/resources_screen.dart
```diff
@@ -169,8 +172,15 @@ class _ResourcesScreen extends State<ResourcesScreen> {
     return AbsorbPointer(
       absorbing: downloading,
       child: ElevatedButton(
-        onPressed: () {
-          state.loadResources(downloadMissing: true);
+        onPressed: () async {
+          try {
+            await state.loadResources(downloadMissing: true);
+          } on SocketException {
+            await state.clearCache();
```
anhappdev, line 179: The app shouldn't delete all the downloaded files when there's no internet connection. If the file being downloaded is corrupted, it should only delete that file. We have checksum validation for every file, so we can identify the corrupted file when the validation fails.

anhappdev (dismissed):

flutter/lib/ui/settings/resources_screen.dart
farook-edev: I cleared the cache in order to reset the download UI, which wouldn't reset on exception. I resolved this now. Thanks for pointing it out!

anhappdev:
@farook-edev We still encounter cases where a downloaded file becomes corrupted, primarily due to connection drops. In such instances, the app should automatically delete the corrupted file. If you prefer, we can create another pull request for this feature.

freedomtan (approved):

flutter/lib/benchmark/state.dart
```diff
@@ -143,15 +143,21 @@ class BenchmarkState extends ChangeNotifier {
...
+      await resourceManager.handleResources(
+        resources,
+        needToPurgeCache,
+        downloadMissing,
+      );
+      print('Finished loading resources with downloadMissing=$downloadMissing');
+      error = null;
+      stackTrace = null;
+      taskConfigFailedToLoad = false;
+    } on SocketException catch (e, s) {
```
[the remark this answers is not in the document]
freedomtan: it's done

PR #960, opened 2025-02-10
feat: download resources per task

anhappdev:
User can download resources per task / benchmark.


freedomtan:
this looks to be desirable. Let's test it. @AhmedTElthakeb and @mohitmundhragithub

freedomtan (approved):

PR #978, opened 2025-03-10
feat(stable-diffusion): added support for models with different quantisation

RSMNYS:
added support for models with different quantisation

@anhappdev  here are the models, so you can copy them: https://drive.google.com/drive/folders/1CT6VUWwGaTw34Za6dTJE7ptms20NUKH3?usp=share_link


freedomtan:
please rebased and get rid of unnecessary changes.

PR #981, opened 2025-03-13
feat: added support for models with different quantization

RSMNYS:
added support for models with different quantization


freedomtan (approved):

freedomtan:
@RSMNYS: please merge this one.

Issue #1013, opened 2025-06-10
make the first-time download flow more user-freindly.


freedomtan:
@farook-edev please help improve first-time user experience (tell the user why she/he has to download resources and go to the resources page first?)

farook-edev:
> Also, what is the point of the "performance" and "accuracy" checkmarks for each model on the Resources page? Can users download accuracy test assets, and if so, how?

@swasson488 This part is a bit confusing 😅, the checkmarks are actually clickable buttons that show the exact files and their availability status (grey check for not downloaded, green for downloaded).

PR #1016, opened 2025-06-24
Feat simple resource flow

farook-edev:
This PR simplifies the resource downloading flow by removing the list of missing files from the warning dialog and making the "Download" button automatically start the download for necessary resources (instead of just going to the resources page).

It also changes the text of the dialog and the location of the "download all" button in accordance with #1013.

This PR is a bit rushed because of Wednesday's release, therefore I'll keep the related issue and branch, and experiment a little bit with making the flow even more straightforward after this is merged.


farook-edev:
@anhappdev could you please check why the windows build is failing? I don't have access to google cloud.

freedomtan (approved):

PR #1017, opened 2025-07-01
chore: Increase targetSdkVersion from 34 to 35

farook-edev:
Similar to #896, PlayStore sent an email to every developer not targeting API 35.


freedomtan:
@anhappdev please help check failed ones.

PR #1018, opened 2025-07-03
fix: Browser stack failures on devices with Qualcomm Soc

Aswinoss:
* Avoids preliminary crash if the delegate is not available
* Adding isFatal check in Str2Delegate function


freedomtan:
@freedomtan to check MLCommons guys for how to make the CLA checking work.

Issue #1055, opened 2025-09-30
implement checking if the app is update-to-date


freedomtan:
@anhappdev please help this.

farook-edev:
I found a couple stackoverflow discussions around figuring out [how the app was installed (play store or not)](https://stackoverflow.com/questions/37539949/detect-if-an-app-is-installed-from-play-store), and [how to get the version number on the play store](https://stackoverflow.com/questions/34309564/how-to-get-app-market-version-information-from-google-play-store) without using the API. For github we can literally just use [the rest API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28&versionId=free-pro-team%40latest&productId=rest).

I hope this info helps!

PR #1061, opened 2025-10-05
feat: prompting users to upgrade when newer app version available

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/1055

Screenshot_1759668660

If users tap on "Update Now," they will be redirected to the app store page of the app, where they can tap on the update button to update the app.


farook-edev (commented):
This PR only works with play store. Users who download the app from github will still be prompted to go to the play store for updates.

farook-edev:
> Is there any better option available?

`package_info_plus` provides [this parameter](https://pub.dev/documentation/package_info_plus/latest/package_info_plus/PackageInfo/installerStore.html) which tells you the ID of the store that installed the app. We can check if it's `com.android.vending` as mentioned [here](https://stackoverflow.com/questions/37539949/detect-if-an-app-is-installed-from-play-store) to confirm whether or not the app was installed via the PlayStore.

farook-edev (commented):
Other than the couple items I noticed in the comments below, LGTM 🚀👍

farook-edev:
wouldn't it be best to merge #1064 and then merge master into this, rather than merging #1064 into this?

PR #1040, opened 2025-08-26
Feat llm

mohitmundhragithub:

PR #1066, opened 2025-10-27
Migrate from Bazel v6.3.2 to v8.4.2

anhappdev:
* Bazel v8.4.2 https://github.com/bazelbuild/bazel/releases/tag/8.4.2 has fixed the issue with macOS Tahoe.
* We currently use Bazel 6, but its end of support is December 2025. Therefore, we should switch to Bazel 8 now. (https://bazel.build/versions/8.4.0/release)


farook-edev:
This probably conflicts with #1040 to some extent. Whichever is merged first will require merging master into the other and some adjustments (and resolving some merge conflicts) will be required.

Just an FYI.

PR #1028, opened 2025-07-21
Resource UI edits

farook-edev:
changed config page and resource page UI to be simpler and more straightforward

closes #1013

PR #1071, opened 2025-11-11
Results for Token throughput based benchmarks

farook-edev:


freedomtan (approved):
LGTM.

freedomtan:
- Should we add the unit for each numbers explicitly on the screen or we can leave them to tooltips when scores are clicked.
- - we are more adding the unit explicitly on the screen, let's think where to add the units. e.g., make "task name" clickable, and add the unit the i column. 
- we should change the history screen, too

farook-edev:
I've added the units next to the result values. As for the 🛈 icon, I was going to make multiple versions and see which one looked best, but then I noticed that the 'make the icon a clickable button and put the info there' idea has already been implemented and approved for the config screen as per #1028. So I opted to replicate the same UI in the result screen.

I also edited the history screen to have both QPS average for non token benchmarks and TPS average for token benchmarks. In addition to adding T/sec or Q/sec for each benchmark in the table below the summary.

<p float="left">
  
  
  
</p>

freedomtan:
@farook-edev @freedomtan: let's check the loadgen source code to make sure that the time of the first token is excluded from the average decoding time.

Issue #1063, opened 2025-10-06
Reduce CI run cost by using merge queue


anhappdev:
Currently, our CI checks run on every new push to a pull request. However, some checks, particularly the integration tests, take a long time and incur higher costs. To address this issue, we can optimize the process by running the tests only before merging into the main branch, rather than on every push.

One way to achieve this is by utilizing a merge queue.
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

PR #1075, opened 2025-11-13
Add merge_group trigger to be able to use merge queue

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/1063

Currently, our CI checks run on every new push to a pull request. However, some checks, particularly the integration tests, take a long time and incur higher costs. To address this issue, we can optimize the process by running the tests only before merging into the main branch, rather than on every push, by using a merge queue.


freedomtan (approved):
LGTM

PR #1029, opened 2025-07-22
chore: upload native debug symbol

anhappdev:
* Closes https://github.com/mlcommons/mobile_app_open/issues/1021
* This PR adds the native debug symbol to the AAB file, which is then uploaded to the Play Store.

PR #1081, opened 2025-11-25
fix: resolve crash due to permission denied on Android Play Store version

anhappdev:
Extracts native libraries from the APK to the file system by setting `useLegacyPackaging true`

https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/dsl/JniLibsPackagingOptions#uselegacypackaging

https://developer.android.com/build/releases/past-releases/agp-4-2-0-release-notes#compress-native-libs-dsl

We previously had this option enabled with `android.bundle.enableUncompressedNativeLibs=false`, but I accidentally removed it during the AGP version upgrade in PR #1029.
```
> Failed to apply plugin 'com.android.internal.application'.
   > The option 'android.bundle.enableUncompressedNativeLibs' is deprecated.
     The current default is 'true'.
     It was removed in version 8.1 of the Android Gradle plugin.
```

### DIFF
--- a/.github/workflows/android-build-test-linux.yml
+++ b/.github/workflows/android-build-test-linux.yml
@@ -16,18 +16,19 @@ env:
   BROWSERSTACK_CREDENTIALS: ${{ secrets.BROWSERSTACK_CREDENTIALS }}
 
 jobs:
-  compute-build-number:
+  computed:
     runs-on: ubuntu-22.04
-    outputs:
-      build_number: ${{ steps.set.outputs.build_number }}
     steps:
       - name: Compute build number
         id: set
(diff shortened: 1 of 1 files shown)


anhappdev:
@freedomtan @Mostelk Please test the Play Store app "593 (5.0.3)" at this link:
https://play.google.com/apps/internaltest/4700449654979131232

freedomtan:
> @freedomtan @Mostelk Please test the Play Store app "593 (5.0.3)" at this link: https://play.google.com/apps/internaltest/4700449654979131232

YES, it works as expected. @Mostelk please check it your devices

freedomtan (approved):

Issue #1074, opened 2025-11-11
Analyst reports crashes on multiple devices


swasson488:
I have been talking to an analyst who would like to use MLPerf mobile in his testing. He just flagged two issues for me:

1) The OnePlus 15 (based on the Snapdragon 8 Elite Gen 5) is crashing constantly when he tries to use the app for testing.

2) The Pixel 10 Pro Fold is crashing in the Stable Diffusion test.

A few questions:

A) Do we expect these devices to have issues?

B) Can we resolve them?

C) Do we expect any new SoC without explicit support in the app to crash? Can we not trap for these issues, avoid crashes, and either fall back to a slower path or provide a clear error message for the user?

anhappdev:
> 2. Pixel 10 Fold issue. [@anhappdev](https://github.com/anhappdev) maybe we should just disable Stable Diffusion in the Pixel backend.

Yes. I think we should do that. I will send a PR soon.

Mostelk:
It crashes on most commercial devices when downloaded from PlayStore, but not when downloaded from github!
Yes, we need extend BrowserStack subscription to test app from PlayStore @swasson488

freedomtan:
Let's discuss with @swasson488 on Wednesday meering

freedomtan:
samsung: Android property, `ro.hardware` 
qualcomm: use opengl GPU id, https://github.com/mlcommons/mobile_app_open/blob/master/mobile_back_qti/cpp/backend_qti/soc_utility.cc#L349-L369
mediatek: https://github.com/mlcommons/mobile_app_open/blob/master/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc#L105, check "ro.hardware" and other software setting, https://github.com/mlcommons/mobile_app_open/blob/master/mobile_back_tflite/cpp/backend_tflite/tflite_c.cc#L53

hardware_check () {
/*
  if (is_q_hardwared()) {
     q_backend();
  } else if (is_samsung_hardware()) {
     s_backend();
  } else {
  }
*/

 if (is_q() and is_samsung()) {
    error("something wrong"); 
 }
}


@anhappdev let's check if we can call the `mlperf_matches_hardware()` for each backend first.

PR #1088, opened 2025-12-16
fix: improve backend matching logic to handle multiple matches

anhappdev:
As commented in https://github.com/mlcommons/mobile_app_open/issues/1074#issuecomment-3639472016


anhappdev:
The 2 devices matched both QTI and Samsung backends:
- qti-Samsung Galaxy S25-15.0
- qti-Samsung Galaxy S25 Ultra-15.0

```
Exception: multiple matching backends found: libqtibackend, libsamsungbackend
```

freedomtan:
> @anhappdev I ran the app and got no immediate errors, although I did get SNPE for everything except stable diffusion. I'm using an SM-S931B/DS which uses a SnapDragon SM8750-AC SOC. 

YES, for stable diffusion, the QNN runtime is used. That's what is designed.

Issue #1058, opened 2025-10-01
LLM Dataset Implementation


farook-edev:
This issue is a general container for matters relating to datasets in general. Discussions on TinyMMLU or IFEval specifically should go in the sub issues for this one.

### TinyMMLU

- [x] Dataset is converted from `.parquet` to `.tfrecord` via a utility script.
- [x] Dataset loads `.tfrecord` and stores data inside samples.
- [x] Dataset provides samples by id to driver/backend in proper format.*
- [x] Dataset Processes output from driver/backend.*
- [x] Dataset calculates and provides accuracy using output data on device.

### IFEval

- [ ] Dataset is converted from `.jsonl` to `.tfrecord` via a utility script.
- [x] Dataset loads `.tfrecord` and stores data inside samples.
- [ ] Dataset provides samples by id to driver/backend in proper format.*
- [x] Dataset Processes output from driver/backend.*
- [x] Dataset calculates and provides accuracy using output data on device.

\* This includes tokenization/detokenization using common SentencePiece utility code.

farook-edev:
- [ ] Datasets should provide token limits as an input

anhappdev:
@farook-edev Can you please share the dataset? So I can upload it to our CDN so that the app can download it automatically.

farook-edev:
> [@farook-edev](https://github.com/farook-edev) Can you please share the dataset? So I can upload it to our CDN so that the app can download it automatically.

Sorry for the delay, here are the dataset and tokenizer files: https://drive.google.com/drive/folders/1nifK2bpi0iXBnG83kDmij6Hip4B_bBIW?usp=sharing

anhappdev:
I think 2 separate benchmarks is a good choice.

PR #1093, opened 2026-01-06
IFEval evaluation code overhaul

### LINKED ISSUE #1060: LLM IFEval Dataset Implementation
Things left to do on this dataset are as follows:

- [x] Finish code to convert data to `.tfrecord`
- [x] Finish code to handle sending and receiving data.
- [x] Review Instruction validation code (possibly against existing implementation).
- [x] Test Accuracy and Performance metrics.
- [x] Potentially move accuracy calculation away from output processing (similar suggestion in TinyMMLU)

farook-edev:
This is an overhaul of the IFEval implementation based on comparisons to google's implementation which uses python. The following are the changes introduced in this PR:

-  `repeat_prompt` instruction now checks to make sure the prompt occurs at the very start of the response.
- `two_responses` instruction is now more flexible, allowing the response to contain the delimiter as an indicator rather than strictly a delimiter.
- `NumberPlaceholders` allows for multi-word placeholders.
- `ConstrainedResponse` allows for the response to not strictly be one of the three options (allowing commas, dots, newlines, etc...).
- `JsonFormat` allows for markdown ticks.
- `MultipleSections` now requires a section number after the word "SECTION", but allows roman numerals.
- `MultipleSections` accounts for the 2 response instructions (where the model will put 2n sections total across the responses).
- `NumberBulletLists` accepts markdown style (`-`) bullets along with star (`*`) bullets.
- `KeywordFrequency` uses stemming ([Snowball based](https://github.com/Blake-Madden/OleanderStemmingLibrary)) and a plural word map to count plural words along with variations of the keyword (e.g. for a keyword `war`, `**war**s` will be allowed but `for**war**d` will not).
- `NthParagraphFirstWord` counts the word itself without any punctuation marks that may be attacked.
- `NumberParagraphs` doesn't miscount separators at the start or end of the response.
- `NumberSentences` properly excludes characters such as dots from being counted as sentence enders if they're numerators or part of an abbreviation etc...
- `EndChecker` is slightly more flexible to account for things like punctuation marks.

With these changes, the reported accuracy from this implementation is considerably more reliable than the reference implementation that Google provides. Merging this PR should finally close #1060 and #1058.

A python version of this implementation is also complete and has been submitted [here](https://github.com/mlcommons/mobile_open/pull/39).


freedomtan (approved):
LGTM. @farook-edev please provide documentation on how to run both C++ and Python code, and compare their outputs, so that we can try to reproduce what you verified.

PR #1094, opened 2026-01-06
Config based LLM pipeline thread count

farook-edev:
This PR uses the logic from `SingleModelPipeline` to assign the thread count in the TFLite config to the `LLMPipeline`.


freedomtan (approved):
LGTM

PR #1095, opened 2026-01-06
Feat. Benchmark Sets

### LINKED ISSUE #1082: Allow more than 1 LLM benchmark
with multiple models such as 1b, 3b, or 8b (and possibly multiple datasets such as mmlu or ifeval)

farook-edev:
This PR adds a new element (BenchmarkSet) which bundles together benchmarks that are mostly similar but need to be run separately (i.e. different models or datasets but same function).

Under the hood the benchmarks work exactly the same, no C++ logic has been changed. The added configuration is only for the frontend.

The way it works is by bundling similar benchmarks under a set, and having each benchmark be active if one or more options it requires are active. For example, if we take LLM, let's say we have 3 models and 3 dataset implementations to test, `ModelA-DatasetB`, `ModelC-DatasetA`, and so on, that'll be 9 benchmarks.
Benchmark `ModelA-DatasetC` will define 2 required options, `Model-A` and `Dataset-C`, then the Benchmark Set will contain 6 options in 2 categories, Models (A,B,C) and DataSets (A,B,C).
If a user then enables Models A and C, And dataset A. the set will automatically activate `ModelA-DatasetA` and `ModelC-DatasetA` and disable all the others.

The benefit from this approach is that instead of having 9 benchmarks that are basically the same, we'll have 1 set containg 6 options. While the core benchmarking will not see the sets or options.

This PR also applies the above described implementation to `image_classification_v2`, combining the default and offline versions into a set, and providing 2 options to enable and disable the benchmarks. This is only a secondary improvements, since this system is meant to tidy up the (at least) 4 benchmarks that LLM will add.

I've also included a video of the system in action:

https://github.com/user-attachments/assets/9c833086-60fc-4d6f-a5bd-bf1bb10cab0a

Closes #1082

### DIFF
--- a/.github/workflows/android-build-test-linux.yml
+++ b/.github/workflows/android-build-test-linux.yml
@@ -271,6 +271,23 @@ jobs:
           path: output/android-apks/*-qsmgt-${{ needs.computed.outputs.build_number }}.aab
           retention-days: 28
           if-no-files-found: error
+      - name: Build Android CLI binaries
+        env:
+          WITH_TFLITE: 1
+          WITH_PIXEL: 1
+          WITH_MEDIATEK: 1
+          WITH_QTI: 1
+          WITH_SAMSUNG: 1
+          WITH_APPLE: 0
+        run: |
+          make cmdline/android/bins/release
+      - name: Archive Android CLI binaries
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: android-cmdline-${{ needs.computed.outputs.build_number }}
+          path: output/android-cli/*
+          retention-days: 28
+          if-no-files-found: error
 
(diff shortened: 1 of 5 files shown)


freedomtan:
@freedomtan to check if the running order (offline one should be the latest one to run) is kept unchanged.

mobile_back_tflite/cpp/backend_tflite/backend_settings/tflite_settings_android.pbtxt
```diff
@@ -36,7 +36,7 @@ benchmark_setting {
 }
 
 benchmark_setting {
-  benchmark_id: "image_classification_v2"
+  benchmark_id: "image_classification_online_v2"
```
anhappdev: We collect run results shared by users. This change will cause a migration if we analyse the results over time.

PR #1097, opened 2026-01-19
MMLU + IFEval Edits

farook-edev:
This PR changes the behavior for inputs larger than the specified limit. It now truncates the first `n` ~tokens (`n = input_size - input_limit`)~ shots until input is within limit, instead of not adding the sample altogether.

I also added the code to connect IFEval to the app and document the cmdline part as per #1098


freedomtan (approved):

freedomtan:
> The TFLite App Crashed on S22 and S25 Ultra

@Mostelk 
It's likely to be memory pressure issue. I tested on some older commercial phones. All failed. However, on some old internal devices with large dram unused, it worked.

@farook-edev and @anhappdev any comments on reducing memory usage?

PR #1117, opened 2026-03-17
LLM raw output

farook-edev:
This PR makes it so that MMLU and IFEval datasets produce the raw output from the model.


freedomtan (approved):
LGTM

PR #1124, opened 2026-04-14
Add toggle all and reset for benchmark sets

### LINKED ISSUE #1122: Make all tasks unchecked by default
With the new collapsible menu, everytime you load the app, Image Classification online and Llama 1B are enabled by default, and they are hidden in the collapsible menu, so user cannot know they are checked!

Can we make all task unchecked by default whenever you start app or  change the mode submission/performance/quick?
It will also be good to add a slider on/off for the Collapsible menus Image Classification and LLM, so it is easy to disable all child tasks which may not be visible from main screen

anhappdev:
## Summary
- Closes https://github.com/mlcommons/mobile_app_open/issues/1122
- Add support for toggling all benchmarks within a benchmark set on/off
- Add ability to reset benchmark states to their defaults
- UI controls added to the benchmark config section and settings screen

## Test plan
- [ ] Verify toggling all benchmarks in a set enables/disables them correctly
- [ ] Verify reset restores benchmark states to defaults
- [ ] Test on both Android and iOS


anhappdev:
The [tflite-Samsung Galaxy M32-11.0 (unified)](https://github.com/mlcommons/mobile_app_open/actions/runs/24432542529/job/71382295311?pr=1124#logs) failed due to a check for an empty checksum in the model configuration.

PR #1143, opened 2026-05-28
Restore MD5 checksums for v6.0 LLM tflite models and disable LLM benchmark for M32 device

anhappdev:
The empty model_checksum entries (introduced in #1113 to ease testing with custom models) caused the integration test validateSettings() to fail for every benchmark, since it iterates allBenchmarks regardless of which benchmark is under test.

Checksums computed from the published model files:
- llama_q8_ekv3072.tflite    -> c618e9bbb89ce52eedcab4f61b2dc3a4
- llama_3b_q8_ekv3072.tflite -> 6bde54c22e2be27bdad5303a0bd9cf72
- llama_8b_q8_ekv3072.tflite -> ec19cf40f881684f15db0b862883ff00

Disable the LLM benchmark for the Samsung Galaxy M32 used in CI tests as it's not supported.

### DIFF
--- a/mobile_back_tflite/cpp/backend_tflite/BUILD
+++ b/mobile_back_tflite/cpp/backend_tflite/BUILD
@@ -28,6 +28,7 @@ pbtxt2header(
     name = "tflite_settings",
     srcs = [
         "backend_settings/tflite_settings_android.pbtxt",
+        "backend_settings/tflite_settings_android_m32.pbtxt",
         "backend_settings/tflite_settings_apple_iphone11.pbtxt",
         "backend_settings/tflite_settings_apple_iphone12.pbtxt",
         "backend_settings/tflite_settings_apple_iphoneX.pbtxt",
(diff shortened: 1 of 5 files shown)


freedomtan (approved):
LGTM

anhappdev:
@freedomtan I think we should disable LLM benchmarks for tflite-Samsung Galaxy M32-11.0.

freedomtan:
> @freedomtan I think we should disable LLM benchmarks for tflite-Samsung Galaxy M32-11.0.

@anhappdev, that sounds like a good idea. Could we add a new configuration for TFLite that disables LLM benchmarks for devices with insufficient DRAM?

anhappdev:
> > @freedomtan I think we should disable LLM benchmarks for tflite-Samsung Galaxy M32-11.0.
> 
> @anhappdev, that sounds like a good idea. Could we add a new configuration for TFLite that disables LLM benchmarks for devices with insufficient DRAM?

@freedomtan Done.

PR #1148, opened 2026-05-29
Resolve conflict for merging submission-v6.0 into master

anhappdev:
**Conflict resolution:**
- `flutter/pubspec.yaml`: kept submission's `6.0.0+1` over master's `5.0.5+1`.
- `WORKSPACE`: kept submission's TF 2.18.0 patch list. Dropped master's `ndk_25_r14.diff` (TF-2.14-only; removed on this branch during the TF 2.18 upgrade), but kept `xcode26_compat.patch` — it only patches `tensorflow/lite/kernels/elementwise.cc`, whose `std::abs<T>` usage is identical in TF 2.18, so it applies cleanly and fixes the Clang 21 `no matching function for call to 'EvalImpl'` error.


farook-edev (approved):
I looked at the parts not related to IOS (since I didn't really touch IOS in the 6.0 branch).

The only concern I have is the Bazel Upgrade to 6.6.0. According to [the Bazel changelog](https://github.com/bazelbuild/bazel/raw/refs/heads/master/CHANGELOG.md), there are some breaking changes between `v6.3.2` and `v6.6.0`.

Most of my concerns are mitigated by CI succeeding, but I'd still keep an eye out, primarily for very new Apple devices.

Issue #1144, opened 2026-05-28
Check merged v6.0 APK


Mostelk:
@mohitmundhragithub @PeggieLu @AhmedTElthakeb  please check the merged v6.0 apk and provide your comments

https://github.com/mlcommons/mobile_app_open/actions/runs/26488614932/artifacts/7231521427)

AhmedTElthakeb:
I tested for samsung, observed two issues. 

1. Stable Diffusion: sd_dec.nnc fails download, says Checksum validation failed .. please download missing files again. 
2. LLM benchmark entry shows up in results screen (despite not supported by backend)

mohitmundhragithub:
it seems to work fine for us.

UI issues are there, which has already been discussed.

freedomtan:
@anhappdev please help check the checksum issue mentioned by @AhmedTElthakeb 

@AhmedTElthakeb for LLM entry, do you mean something like this

<img width="120"  alt="Image" src="https://github.com/user-attachments/assets/14ab1497-6035-4e22-a04e-1c43ca634f88" />

anhappdev:
> Stable Diffusion: sd_dec.nnc fails download, says Checksum validation failed .. please download missing files again.

@AhmedTElthakeb, please clear the cache and download again. I checked, and the sd_dec.nnc on the server has the same checksum (`e820be61269dedb1b7f91efcf9b5c4d0`) defined in the task file.

freedomtan:
@PeggieLu, how about the Dimensity 9500? Does the unified app work as expected?

Mostelk:
> [@anhappdev](https://github.com/anhappdev) please help check the checksum issue mentioned by [@AhmedTElthakeb](https://github.com/AhmedTElthakeb)
> 
> [@AhmedTElthakeb](https://github.com/AhmedTElthakeb) for LLM entry, do you mean something like this
> 
[...]

PR #1151, opened 2026-06-02
Remove empty sets from results screen

farook-edev:
This PR fixes the issue in #1144 regarding phantom llm entries.

It uses the same logic as in the config section.


freedomtan (approved):
LGTM. Yes, the phantom LLM entry on Pixel 10 is removed. @Mostelk and @mohitmundhragithub you may want to test [this unified apk](https://github.com/mlcommons/mobile_app_open/actions/runs/26813667734/artifacts/7355713301).

PR #1164, opened 2026-07-30
feat: per-benchmark backend selection (mix CoreML and TFLite per benchmark)

anhappdev:
## What

Per-benchmark backend selection. Previously one backend was chosen per session; on iOS with `WITH_APPLE=1`, CoreML always won and benchmarks it doesn't declare (`llm-*`, `stable_diffusion`) silently disappeared.

- All backends matching the device are offered (`findMatchingBackends()`), in `list.in` priority order. Each benchmark defaults to the highest-priority backend that supports it, with a per-benchmark backend dropdown in the config screen (mirroring the delegate dropdown); delegate choices, model files, and batch size follow the selection. Single-backend devices render the same static label and behavior as before.
- Whether the TFLite fallback is offered next to a vendor backend is controlled by the vendor via `BackendSetting.fallback_policy` in its own settings file. The default `FALLBACK_DISABLED` reproduces the pre-PR behavior (vendor backend only, undeclared benchmarks hidden); `FALLBACK_FILL_GAPS` lets TFLite serve only benchmarks the vendor doesn't declare; `FALLBACK_COEXIST` offers TFLite as a selectable alternative everywhere. The Core ML backend opts into `FALLBACK_COEXIST`, so on iOS LLM benchmarks appear on TFLite while image classification defaults to CoreML. Android vendor devices behave exactly as before unless the vendor opts in.
- Explicit backend choices persist across restarts; exports record the backend that actually ran, from the run's own settings snapshot.

No C++ or result-schema changes — the native bridge already loads the backend lib per run; the only proto change is the additive `fallback_policy` field. `BackendInfoHelper.alwaysOfferFallback = false` restores the historical behavior globally in one line.


freedomtan:
Let's enable this feature on for the CoreML backend for now.

We can enable it when LiteRT + GPU accelerator (and NPU accelerator, maybe) is merged.

anhappdev:
@freedomtan For iOS you can test the build 255. For Android you can use the APK built by the CI in this PR.

freedomtan (approved):
LGTM

PR #1165, opened 2026-08-02
fix: let iOS builds without Firebase secrets (use dummy values) + @main migration

anhappdev:
## What

Lets iOS builds without Firebase secrets launch. The app is designed to run Firebase-free (`FirebaseManager.enabled`), but without secrets the generated `GoogleService-Info.plist` contained empty values, and firebase-ios-sdk kills the app at launch anyway: `FIRApp` rejects an empty/malformed `GOOGLE_APP_ID`, and `FirebaseInstallations` separately validates `APIKey` (exactly 39 chars, leading `A`, base64 url-safe) and `projectID` at app activation — even when Dart never initializes Firebase.

- `generate-firebase-config-files.sh` now defaults `FIREBASE_IOS_APP_ID`, `FIREBASE_IOS_API_KEY`, and `FIREBASE_PROJECT_ID` to well-formed dummies that pass all native validation. Real env values (CI secrets) take precedence.
- `DefaultFirebaseOptions.available()` recognizes the dummy project ID, so Firebase stays disabled on the Dart side.
- Also migrates `AppDelegate` from the deprecated `@UIApplicationMain` to `@main` (the Flutter 3.44 tool auto-applies this on every run, dirtying each developer's tree).

### DIFF
--- a/flutter/ios/Runner/AppDelegate.swift
+++ b/flutter/ios/Runner/AppDelegate.swift
@@ -1,7 +1,7 @@
 import Flutter
 import UIKit
 
-@UIApplicationMain
+@main
 @objc class AppDelegate: FlutterAppDelegate {
   override func application(
     _ application: UIApplication,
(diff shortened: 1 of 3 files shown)


freedomtan (approved):
LGTM

PR #1166, opened 2026-08-11
Allow benchmark-set option changes to persist after app closing

### LINKED ISSUE #1122: Make all tasks unchecked by default

### LINKED ISSUE #1140: Task set options should save their state
Similar to how normal benchmarks work, benchmark set options should save their state when changed, so that when the app is reopened, selected options do not reset.

farook-edev:
This PR makes it so that enabled or disabled options remain that way after closing and re-opening the app.

Closes #1140, #1122, and possibly supersedes #1124

### DIFF
--- a/flutter/lib/benchmark/benchmark.dart
+++ b/flutter/lib/benchmark/benchmark.dart
@@ -161,6 +161,16 @@ class BenchmarkSet {
     return availableOptions().length;
   }
 
+  Map<String, bool> optionStateMap() => {
+    for (BenchmarkOption item in availableOptions()) item.id: item.enabled,
+  };
+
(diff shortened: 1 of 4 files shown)


freedomtan (dismissed):
LGTM

anhappdev:
@farook-edev Can you please resolve the merge conflict?

anhappdev (approved):

Issue #1168, opened 2026-08-12
noticed a new warning while buiding the app


mohitmundhragithub:
>mappingFileUploadEnabled and nativeSymbolUploadEnabled set to false
WARNING: Your Android app project: app located at: /image-workdir/project/flutter/android/app/build.gradle
applies the Kotlin Gradle Plugin, which will cause build failures in future versions of Flutter. 
Please migrate your app to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers
\
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): desktop_webview_auth, firebase_app_check, firebase_storage, wakelock_plus
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.
\
Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing 
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors
\
If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

Started seeing this warning while building the app.

PR #1170, opened 2026-08-15
chore: stop applying KGP (Kotlin Gradle Plugin) in the app, upgrade wakelock_plus to 1.7.0

anhappdev:
Silences the Kotlin Gradle Plugin warnings Flutter 3.44 prints on every Android build, and removes one of the four plugins responsible.

### What changed

- Dropped `id "kotlin-android"` from `flutter/android/app/build.gradle`. The Flutter Gradle Plugin applies it itself now, and the 3.44 `flutter create` template no longer lists it.
- `settings.gradle` is untouched, so KGP stays pinned at 2.2.20.
- Upgraded `wakelock_plus` 1.6.1 → 1.7.0, which migrated to Built-in Kotlin. Lockfile only — `^1.2.8` already allowed it.

### The four flagged plugins

| Plugin | Status |
| --- | --- |
| `wakelock_plus` | Fixed here |
| `firebase_app_check`, `firebase_storage` | False positives — already guarded behind `agpMajor < 9 \|\| !builtInKotlin`; Flutter's regex detector can't see the guard |
| `desktop_webview_auth` 0.0.16 | No fix exists; transitive via `firebase_ui_auth` → `firebase_ui_oauth`. Needs an upstream issue |

### Review notes

- The full migration (`android.builtInKotlin=true`, `kotlinOptions` → `kotlin.compilerOptions`) needs Flutter 3.47+ / AGP 9. Deferred to the next toolchain bump.
- `:app:kgpVersion` still reports 2.2.20, so we did not silently fall back to the 2.0.0 KGP bundled with the Flutter Gradle Plugin.
- `:app:compileDebugKotlin` is still in the task graph, confirming `kotlin-android` is applied.
- CI green: Build Android app, build-android-apk, both iOS builds, and the on-device runs all pass.

### DIFF
--- a/flutter/android/app/build.gradle
+++ b/flutter/android/app/build.gradle
@@ -1,6 +1,9 @@
 plugins {
     id "com.android.application"
-    id "kotlin-android"
+    // Do not apply "kotlin-android" here. The Flutter Gradle Plugin applies KGP
+    // itself, using the version pinned in settings.gradle. Applying it here makes
+    // Flutter warn about migrating to Built-in Kotlin and will break future builds.
+    // https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers
     id "dev.flutter.flutter-gradle-plugin"
     id 'com.google.gms.google-services'
     id 'com.google.firebase.crashlytics'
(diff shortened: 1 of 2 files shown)


freedomtan:
this is supposed to resolve some of the new warnings reported by @mohitmundhragithub, #1168 
@mohitmundhragithub please check this.

freedomtan (commented):
LGTM.

@mohitmundhragithub you may want to take a look

freedomtan (approved):
LGTM

PR #1171, opened 2026-08-18
Add mobile_back_litert: LiteRT compiled-model backend

anhappdev:
Adds `mobile_back_litert`: an Android-only backend that runs every benchmark on the LiteRT 2.1.5 CompiledModel API - the LLM benchmarks on a dedicated LLM pipeline, stable diffusion on the stable-diffusion pipeline, the vision/NLP benchmarks on a single-model pipeline.

### How it works
- `llm_pipeline` drives a `litert::CompiledModel` with explicit `TensorBuffer`s for prefill and decode.
- `llm-1b` and `llm-1b-instruct` default to the GPU delegate with a dedicated GPU model export (`llama_q8_ekv3072_litert.tflite`). GPU compilation failure falls back to CPU automatically.
- The larger LLM benchmarks run on the CPU delegate.
- `single_model_pipeline` runs the vision/NLP benchmarks on `litert::CompiledModel`. The GPU accelerator is the default (fp32 model exports, automatic CPU fallback); the CPU delegate choice runs the int8 exports on XNNPACK. Inputs and outputs stage through host buffers so the mlperf harness keeps stable pointers.
- `stable_diffusion_pipeline` runs stable diffusion (added in #1174). CPU only: the shipped exports are dynamic_int8 (text encoder, diffusion) and dynamic_fp16 (decoder), aimed at CPU/XNNPACK, so a GPU choice would need dedicated fp32 exports. The sub-models bind by name - the CompiledModel signature order is alphabetical, not positional.
- NNAPI is not offered: it is deprecated since Android 15, and the CompiledModel NPU path needs vendor SDKs plus AOT-compiled models (the Google Tensor SDK is a sign-up-gated beta, G5-only). The LiteRT team's interim advice for Pixel is the GPU delegate. NNAPI-EdgeTPU coverage stays with the pixel backend on the Pixel 9 Pro job.
- The backend is listed before `libtflitebackend`, so LLM defaults to LiteRT; on Pixel devices the pixel backend still outranks it for the vision benchmarks.
- The backend declines to load on the Galaxy M32 (SM-M326B), which does not have the memory for the LLM benchmarks.

### Backend coexistence
- `fallback_policy` is replaced by a `claim_policy` chain: per benchmark, the app walks the backends in priority order and stops after the first claimant that is not `CLAIM_SHARED`.
- Every backend declares `CLAIM_SHARED`, so users can pick any capable backend per benchmark. Defaults still follow the priority order, so CI runs are unchanged.
- This adds one declaration to each vendor settings file - the only vendor-file change in this PR.

### Build and packaging
- LiteRT is an opt-in flavor like the vendor backends: `WITH_LITERT` defaults to 0, and the unified test APK, release APK/AAB, and Android CLI builds request it explicitly.
- APK filename letters: LiteRT adds `l`, and the Pixel letter changes from `g` to `p` (unified suffix is now `qsmplt`).
- The board archives a litert-only test APK (`android-apk-litert-<n>`).
- The Sonar build keeps `WITH_LITERT=1`: this backend is first-party code and stays analyzed, unlike the proprietary vendor SDKs.

### Merged from `compiled_model`
- The single-model pipeline is compiled into the `.so`, ported to the CompiledModel API (see above). The stable-diffusion pipeline followed in #1174, also ported to CompiledModel. The dummy-backend scaffolding is imported but not wired up; the iOS/Windows settings variants and the dead iOS delegate branches were dropped (this backend is Android-only).

### Toolchain changes
- Bazel 7.7.0.
- `@litert` 2.1.5; `org_tensorflow` pinned to the commit LiteRT expects.
- `rules_apple`/`apple_support` pinned to LiteRT's versions; small patches for the pinned TF revision.
- Android builds at NDK API 30 (matches `minSdkVersion`). Building at 33 made absl reference `backtrace()`, which Android 11 devices lack, so the backends failed to dlopen there.

### CI
- Device jobs always test the unified APK; the integration test decides the backend. On the Pixel 10 Pro it pins every benchmark to the LiteRT backend (`deviceBackendOverride`), so that job covers LiteRT vision and LLM (GPU) plus stable diffusion (CPU) end to end.
- The Pixel 9 Pro job keeps the defaults: vision on the pixel backend, `llm-1b`/`llm-1b-instruct` on LiteRT.
- Measured on the Pixel 10 Pro job at `1b9e045` - all nine benchmarks on `liblitertbackend`, `All tests passed`. Accuracy comes from the CI subset in quick mode, so it is only meaningful against TFLite on the same run, not as a real accuracy score.

  | benchmark | accel | throughput | accuracy |
  | --- | --- | --- | --- |
  | image_classification_v2 | GPU | 14.74 | 0.830 |
  | object_detection | GPU | 21.91 | 0.3162 |
  | image_segmentation_v2 | GPU | 19.28 | 0.3921 |
  | natural_language_processing | GPU | 1.757 | 1.000 |
  | super_resolution | GPU | 8.789 | 0.3400 |
  | image_classification_offline_v2 | GPU | 28.31 | 0.900 |
  | llm-1b | GPU | 3.595 tok/s | 1.00 |
  | llm-1b-instruct | GPU | 3.593 tok/s | 0.25 |
  | stable_diffusion | CPU | 0.00796 | perf-only |

- LLM CPU baselines: 4.9-8.0 tok/s (Pixel 9 Pro) and 7.97-8.14 tok/s (Pixel 10 Pro). Expected throughput intervals are still deliberately wide (smoke test, not a perf gate); they can be tightened now that the GPU and LiteRT-vision numbers have landed.
- The unified device jobs got a 90-minute budget and a longer pre-run wait in the test: with the GPU delegate default the LLM benchmarks carry two ~1.2 GB models, doubling the download and checksum-validation time.
- Vendor jobs (qti/samsung/mtk) skip LLM until those backends support it.
- `tf_nnapi_no_mmap_sharing.patch`: the rebuilt TFLite passed the model file's fd to NNAPI drivers. SELinux rejects that for models on external storage, so NNAPI fell back to CPU. The patch restores master's behavior; Pixel EdgeTPU throughput matches master again.
- Windows build moved from Google Cloud Build to GitHub Actions (`windows-build-test.yml`).
- MSVC needed three fixes: hide host Android NDK env from repo rules; LiteRT's `is_msvc = True` rewrite for `xla/tsl`; protobuf's `--define=protobuf_allow_msvc=true`.
- The Linux Android job uses a GCS bazel cache (same pattern as the Xcode Cloud build) instead of actions/cache, whose 10 GB quota was churned by buildkit blobs and whose keys never re-save.
- Workflow downloads are pinned, sha256-verified, and https-enforced.
- iOS builds ride out known flakes: on a remote-cache error the build retries without the cache; `pod install` backs off on CDN 429s.
- The legacy Cloud Build trigger can be disabled once the new workflow is trusted.

### Follow-ups
- #1175 adds iOS support (CPU + Metal) on top of this branch.

### Review notes
- Two changes from 530df217 are reverted as local testing shortcuts: the `llm-*` tasks in `tasks.pbtxt` are restored, and `tokenThroughput` no longer returns `double.infinity`. Please confirm neither was deliberate.


freedomtan:
with current build, we can get the Pixel apk, which comes with LiteRT (CPU) for LLM. 

@farook-edev to share a newly converted 1b model, supposedly, it will work with LiteRT GPU accelerator.
@anhappdev will update the app when the model is uploaded.

freedomtan:
@anhappdev please merge @farook-edev's latest commit from compiled_model. And then we have the same baseline to work on.

freedomtan:
@anhappdev will enable LiteRT for the unified app, we will be able to test the LiteRT backend on any device, hopefully.

Mostelk:
@anhappdev does this PR enable LiteRT on all backends?https://github.com/mlcommons/mobile_app_open/actions/runs/32988027240/artifacts/9615977270

anhappdev:
> @anhappdev does this PR enable LiteRT on all backends?https://github.com/mlcommons/mobile_app_open/actions/runs/32988027240/artifacts/9615977270

@Mostelk In current state all benchmarks (except SD) can be run with LiteRT backend on any Android phones.

anhappdev:
@freedomtan @farook-edev LiteRT can run LLM with GPU now. All benchmarks (except Stable Diffusion) can run with LiteRT (tested on Pixel 10 Pro via BrowserStack).

anhappdev:
### LiteRT on Pixel 10 Pro - measured results

From the `litert-Google Pixel 10 Pro-16.0 (unified)` job (run [32973058131](https://github.com/mlcommons/mobile_app_open/actions/runs/32973058131), head 5d4b7655). The unified APK with `deviceBackendOverride` pinning every benchmark to `liblitertbackend`; all of them ran on the **GPU** accelerator, with no fallback to another backend or to CPU.

| benchmark | throughput | accuracy |
| --- | --- | --- |
| image_classification_v2 | 14.28 QPS | 83.00% |
| object_detection | 22.68 QPS | 0.3162 mAP |
| image_segmentation_v2 | 18.67 QPS | 0.3921 mIoU |
| natural_language_processing | 1.78 QPS | 100.0000 |
| super_resolution | 8.35 QPS | 34.00 dB |
| image_classification_offline_v2 | 27.44 QPS | 90.00% |
| llm-1b | 3.59 tok/s | n/a |
| llm-1b-instruct | 3.54 tok/s | n/a |
| stable_diffusion | not supported | - |

Test result: `30:45 +10: All tests passed!` in 30m46s.

Notes:
- The LLM benchmarks run in `quickRun` mode in the integration test (2 queries against `min_samples: 10`), so their accuracy is not meaningful and loadgen marks the result invalid. Everything else ran the full accuracy dataset.
- Two benchmarks are partially delegated, which is expected: object_detection puts 1 of 125 ops on CPU (`TFLite_Detection_PostProcess` has no GPU kernel) and natural_language_processing puts 10 of 2742 on CPU (`GATHER` supports only 1D indices). Everything else is fully delegated - image classification is 129/129 nodes on `LITERT_CL`, and the LLM graphs are 796/796 (prefill) and 782/782 (decode).
- `image_classification_offline_v2` runs on a single shard (`shards_num: 1`, batch 2): two shards calling `Run()` concurrently on one GPU environment deadlock.
- Stable diffusion is not wired up on this backend yet - the SD pipeline sources are present but not compiled in, and they still use the Interpreter API. Being addressed separately.

freedomtan:
> ### LiteRT on Pixel 10 Pro - measured results
> From the `litert-Google Pixel 10 Pro-16.0 (unified)` job (run [32973058131](https://github.com/mlcommons/mobile_app_open/actions/runs/32973058131), head [5d4b765](https://github.com/mlcommons/mobile_app_open/commit/5d4b765557bc2e68b80a6f736f1c6e59cf40757e)). The unified APK with `deviceBackendOverride` pinning every benchmark to `liblitertbackend`; all of them ran on the **GPU** accelerator, with no fallback to another backend or to CPU.
> 
> benchmark	throughput	accuracy
> image_classification_v2	14.28 QPS	83.00%
> object_detection	22.68 QPS	0.3162 mAP
> image_segmentation_v2	18.67 QPS	0.3921 mIoU
> natural_language_processing	1.78 QPS	100.0000
> super_resolution	8.35 QPS	34.00 dB
> image_classification_offline_v2	27.44 QPS	90.00%
> llm-1b	3.59 tok/s	n/a
> llm-1b-instruct	3.54 tok/s	n/a
> stable_diffusion	not supported	-
> Test result: `30:45 +10: All tests passed!` in 30m46s.
> 
> Notes:
> 
> * The LLM benchmarks run in `quickRun` mode in the integration test (2 queries against `min_samples: 10`), so their accuracy is not meaningful and loadgen marks the result invalid. Everything else ran the full accuracy dataset.
> * Two benchmarks are partially delegated, which is expected: object_detection puts 1 of 125 ops on CPU (`TFLite_Detection_PostProcess` has no GPU kernel) and natural_language_processing puts 10 of 2742 on CPU (`GATHER` supports only 1D indices). Everything else is fully delegated - image classification is 129/129 nodes on `LITERT_CL`, and the LLM graphs are 796/796 (prefill) and 782/782 (decode).
> * `image_classification_offline_v2` runs on a single shard (`shards_num: 1`, batch 2): two shards calling `Run()` concurrently on one GPU environment deadlock.
> * Stable diffusion is not wired up on this backend yet - the SD pipeline sources are present but not compiled in, and they still use the Interpreter API. Being addressed separately.

@anhappdev do you use full validation set or a tiny subset? these numbers are dubious. For example, the natural_language_processing one, 
* 1.78 looks like CPU performance
* 100.0 accuracy is very weird.

anhappdev:
> @anhappdev do you use full validation set or a tiny subset? these numbers are dubious. For example, the natural_language_processing one,
> 
> * 1.78 looks like CPU performance
> * 100.0 accuracy is very weird.

The benchmarks run on CI with quick mode and it use a subset designed for CI so the accuracy is just for comparing with the TFLite backend on the same CI run. I do not have the full dataset to test the real accuracy.

freedomtan:
@farook-edev to add some changes including Gemma 4 ones.

freedomtan:
@farook-edev text encoder, unet/diffusion model (50 iterations), VAE decoder. Let's try to enable the UNET one on GPU accelerator first. It seems no re-conversion is needed. Let's try it.

farook-edev:
gemma4 model files available [here](https://drive.google.com/drive/folders/1VC2TpmmV1wcpqB4nm4EsrNbAZhu8zyte?usp=drive_link)

freedomtan:
@anhappdev 

- [x]  please help check the model checksum issue. 

- [x] please help put gemma 4 models to cloud side and modify the URLs.

@farook-edev

- accuracy numbers of the gemma 4 (dynamic w8, activations stay with fp32) you converted.
  - what does the "dynamic wi8" mean?
  - [x] tinyMMLU is done to certain extent
  - [ ] ifeval, still checking
- [] performance numbers, too
  - slower than llama 1B.
  - let's try to see if we can get some profiling data (comparing with llama 1b).
- [x] things to try:  the QAT model (int8 <-- well-support, int8, fp8, and others (fp4?) <-- ??) 
  - litert-torch doesn't support converting pre-quantized checkpoints.
- let's try to convert QWen 3.5 0.8B: https://huggingface.co/Qwen/Qwen3.5-0.8B
- [x] make sure that the pipe can handle llama 1B, gemma 4 E2B, and QWen 3.5 0.8B ones.
  - the pipeline works for both llama and gemma 4 now.

farook-edev:
I found this on quantization [here](https://github.com/google-ai-edge/ai-edge-quantizer/blob/main/ai_edge_quantizer/recipe.py)
```
"""Quantization recipe module.

Existing recipes include:

  1. Dynamic quantization recipes:
    - dynamic_wi8_afp32
    - dynamic_wi4_afp32
  2. Weight-only quantization recipes:
    - weight_only_wi8_afp32
    - weight_only_wi4_afp32
  3. Static quantization recipes:
    - static_wi8_ai8
    - static_wi8_ai16
  4. LiteRT-LM recipes:
    - gemma4_mixed48
    - gemma4_mixed48_hr
    - gemma4_mixed48_b32
    - gemma4_mixed48_b64

Naming convention decoder for recipes:
  - 'dynamic': dynamic range quantization (weights quantized statically,
    activations dynamically at runtime).
  - 'wi[N]': weight integer N-bit (e.g., wi8: 8-bit weights).
  - 'c' or 'b[M]': Granularity. 'c' for channelwise, 'b[M]' for blockwise with
    block size M (e.g., b32: 32-blockwise).
  - 'hr' (optional): Hadamard rotations are used for the quantization
    parameters estimation. (typically for better quality at lower bits, see
    algorithms/uniform_quantize/hadamard_rotation.py).
  - 'afp32': activations remain float32 in the model (may be dynamically
    quantized at runtime by setting compute_precision=INTEGER,
    explicit_dequantize=False).
"""
```

freedomtan:
If are willing to download models and put them to your device manually, then you can try https://github.com/mlcommons/mobile_app_open/actions/runs/34861220129/artifacts/10356801955. Otherwise, we can wait for the automatically downloading one.

anhappdev:
### Gemma E2B measured on BrowserStack

Both Gemma benchmarks now download from `gs://mlperf-mobile-public/litert/gemma4/` and the checksums verify on device, so the models load and run on real hardware. Everything below is LiteRT / GPU, `quickRun`.

The only run where all four completed is build 1035. Build 1036 (`9588d5ca`) reproduced Pixel 10 Pro `llm-gemma-e2b` at 1.67 tok/s, and its other three were cut short — see the idle-timeout note below.

| device | benchmark | tok/s | TTFT | queries | duration | score |
|---|---|---|---|---|---|---|
| Pixel 9 Pro | `llm-gemma-e2b` | 5.64 | 2.26 s | 6 | 64 s | 0.667 |
| Pixel 9 Pro | `llm-gemma-e2b-instruct` | 63.63 | 2.24 s | 2 | 136 s | 0.125 |
| Pixel 10 Pro | `llm-gemma-e2b` | 1.67 | 3.85 s | 2 | 59 s | 0.0 |
| Pixel 10 Pro | `llm-gemma-e2b-instruct` | 405.64 | 3.60 s | 2 | 421 s | 0.125 |

**These are not performance numbers.** Every run reports `is_result_valid: false` with `is_min_query_met: false` — quickRun only gets 2–6 queries in — and there is no accuracy phase (`accuracy_run` is null), so the score column is the performance run's incidental result over a handful of samples, not an accuracy measurement.

Two things worth a look.

**The instruct throughput is not measuring generation rate.** Same model and device as the non-instruct row, yet 405 tok/s against 1.67. It also rises as the device gets *slower*: Pixel 9 Pro 136 s → 63.6, Pixel 10 Pro 421 s → 405.6. 421 s at 405 tok/s implies ~170k tokens from 2 queries. Something is off in the token accounting on the IFEVAL path.

**The Gemma benchmarks exceed BrowserStack's idle timeout, which is what the two red device jobs are.** `IDLE_TIMEOUT=900` in `.github/workflows/scripts/browserstack-app-automate.sh`, and one Gemma benchmark is a single uninterrupted test-framework call:

- Pixel 9 Pro `llm-gemma-e2b`: 890 s in build 1035 (passed), 894 s in build 1036 (killed)
- Pixel 10 Pro `llm-gemma-e2b-instruct`: 880 s in build 1036 (killed)

Most of that is the 4.7 GB download plus the LiteRT GPU compile (`Replacing 1534 out of 1534 node(s) with delegate (LITERT_CL)`); only ~90–140 s is actual benchmarking. Build 1035 cleared the ceiling by ten seconds, which is why this looks intermittent rather than broken. Note the `keepalive ping` in `flutter/integration_test/first_test.dart` is a `debugPrint` — it reaches logcat but does not reset BrowserStack's idle timer, so it does not help here.

Raising `IDLE_TIMEOUT` would unblock the device jobs, but the download and compile cost is the real constraint.

Mostelk:
@farook-edev is it possible to check pipeline with the Gemma 4 E4B

PR #1172, opened 2026-08-20
[DO NOT MERGE] CI bisect: pixel NNAPI isolation (WITH_LITERT=0)

anhappdev:
Temporary experiment for PR #1171: identical branch with WITH_LITERT=0, so the APK carries no LiteRT libraries but keeps the new TF pin and rebuilt backends. If the pixel job still hits the NNAPI binder failures, the regression is in the shared rebuild; if it passes, it is tied to the LiteRT libraries in the app. Will be closed after one CI run.


anhappdev:
Bisect complete: the pixel NNAPI failure reproduces without any LiteRT library in the APK, and master passes on the same device pool the same day. Root cause: the rebuilt TFLite runtime passes the model file's mmap fd (external storage/fuse) to the EdgeTPU HAL, which SELinux forbids from reading it — binder rejects the fd transfer and delegation falls back to CPU. Fixed on the litert branch by patching out TFLITE_NNAPI_ALLOW_MMAP_SHARING (commit 876a13da).

PR #1173, opened 2026-08-27
fix: make the bazel GCS cache actually hit across runs to reduce build time

anhappdev:
Makes the GCS bazel cache from #1171 actually hit across CI runs, and rations the BrowserStack device tests to the account's 2-session limit.

## bazel cache

Two things stopped the cache from ever hitting:

- **Linux** inherited `PATH` into every action key. `setup-gcloud` puts a per-run `$RUNNER_TEMP/<uuid>` on `PATH`, so every key changed every run — the cache held ~19k entries and hit 7 of 4497 actions.
- **macOS** had three jobs sharing one `actions/cache` key over one path. Only one could write it, so the Android job kept restoring the iOS job's cache and never saved its own.

Fixes:

- `--incompatible_strict_action_env` in `.bazelrc` pins the action `PATH`. Repo rules still see the real environment.
- Pin loadgen's build-date stamps in `WORKSPACE` (the only non-reproducible fetched file).
- Linux uses the GCS remote cache. macOS keeps a per-config disk cache **in front of** GCS (fast local hits, GCS for cold cases).
- Every cache key is now per configuration, so the collision can't recur. Fork PRs with no GCS secret fall back to disk-only.
- Fix the CocoaPods key (was hashing an untracked `Podfile.lock`, collapsed to a constant).

Result — all jobs now hit 100% of cacheable actions (zero local):

| job | before | after |
|---|---|---|
| Linux build | ~67 min | **33 min** |
| macOS Android | 35 min | **23 min** |
| iOS apple / tflite | ~15 min | ~15 min (durability, not speed) |

## BrowserStack

The account allows 2 parallel sessions; runs were failing with `BROWSERSTACK_ALL_PARALLELS_IN_USE`.

- Per-PR `concurrency` group on the four device jobs: a new commit cancels the previous run's job for the same device. Never workflow-wide — builds always finish.
- `max-parallel: 1` on those matrices, so one run can't exceed the budget alone.
- Device-job `timeout-minutes` 60 → 135, so the built-in retry (`max_attempts: 2`) can actually run instead of being killed after the first attempt.

---

Once merged, #1171 can drop its `ci: move the Linux bazel cache to GCS` commit and rebase.


anhappdev:
@freedomtan Please review this one

freedomtan (approved):
LGTM

PR #1174, opened 2026-08-27
Run stable diffusion on the LiteRT backend

anhappdev:
Ports the dormant stable-diffusion pipeline to the LiteRT 2.1.5 CompiledModel API and wires it in, closing the last benchmark gap on this backend. Targets `litert`, not master.

### How it works
- Three models (text encoder, diffusion, decoder) share one `litert::Environment`; 43 invocations per query at the shipped `num_steps: 20`. Buffers are created once and reused across the denoising loop.
- **Inputs are bound by signature name, never by position.** The signature keys are ordered alphabetically, which does not match tensor order: the text encoder reads `(positions, tokens)` and the diffusion model reads `(context, latent, timestep_embedding)`. Binding positionally would swap latent with context - both float32, so it would produce a plausible but wrong image rather than an error. A missing name is a hard failure that logs the names actually found.
- CPU only. The shipped exports are `dynamic_int8` (encoder, diffusion) and `dynamic_fp16` (decoder), aimed at CPU/XNNPACK; a GPU choice would need dedicated fp32 exports. Despite the filenames, nothing is quantized at the graph boundary - encoder inputs are int32, everything else float32. The delegate is labelled `CPU`/`cpu` accordingly, rather than copying the TFLite entry's `NNAPI`/`npu` labels, which are cosmetic (that code attaches no delegate).
- No new model exports and no `tasks.pbtxt` change: the existing v5_0 TFLite SD models are plain flatbuffers and load as-is. All four checksums were verified against the downloaded files.

### Test coverage, and its limit
Stable diffusion joins the LLM benchmarks in `quickRun`. In accuracy mode loadgen ignores `min_query_count` and runs the whole set - here 150 sequential 20-step generations plus 150 CLIP scoring passes and a 1.71 GB ground-truth download, which is hours on a CPU against a 90-minute job. `quickRun` keeps the time-bounded performance phase.

The consequence is worth stating plainly: **CI exercises the pipeline end to end but cannot catch a wrong-but-plausible image**, because nothing scores the output. Name-based binding plus hard failure on a missing name is what guards correctness instead.

The QTI-only gate in `canRunBenchmark` is widened for LiteRT by inverting the condition rather than adding `|| litert`, so QTI behaviour is bit-identical and SD does not start running on the Samsung and MediaTek jobs, which would add ~2.8 GB of downloads and an uncapped CPU run to three more device budgets.

### Defects fixed in the dormant sources
Carried over from the TFLite twin, all pre-existing: `exit(-1)` on inference failure now returns `MLPERF_FAILURE`; the per-query invoker leak is gone; the unbounded scan for a token terminator is bounded at 77 and the token tail zero-padded, so no stale data from a previous sample reaches the model; the cumulative-alpha index is bounds-checked; the embeddings loader validates its reads before allocating. Dead code dropped: `get_tensor_index_by_name`, two unused `get_timestep_embedding` implementations, and a `main()` behind `__TEST_BPE__`.

Note these files were byte-identical to `mobile_back_tflite`'s copies, so they now diverge - the same defects remain in the TFLite backend and are worth a follow-up there.

### Verification
Confirmed on a Pixel 10 Pro (run 33053477420). Stable diffusion ran end to end on `liblitertbackend`: the pipeline initialised, all six tensors resolved by signature name with no binding failure, and both samples completed the full 20-step denoising loop - 111.31 s mean latency per image (111.71 s p90), 0.00895 QPS, 2 samples in 229.9 s. `isResultValid` is false only because the min-query rule (2 vs 128) is unmet under the time-bounded run, exactly as the LLM benchmarks behave.

Two incidental fixes rode along, both surfaced by this work:
- `checkAccuracy` dereferenced `accuracyRun` with a null check operator, which is null in `quickRun`. The LLM benchmarks never hit it only because their expected-accuracy maps are empty and return early. The run mode's own `doAccuracyRun` flag is now threaded through from the call site, so the check is skipped only when the mode genuinely has no accuracy phase - rather than whenever the field happens to be null, which would silently stop checking benchmarks that should have one.
- The BrowserStack trigger retry went from 2 attempts to 8. The device jobs have no `concurrency:` group, so two boards at once exhaust the account's parallel budget and every loser fails outright with `BROWSERSTACK_ALL_PARALLELS_IN_USE` - no session created, nothing run. That cost four device-job failures on this branch that looked like code failures. `retry_on_exit_code: 9` is specifically the trigger failure, so this only lengthens how long a job waits for a slot; everything else still fails fast. A concurrency group would address the cause rather than the symptom, but it changes scheduling for every board and belongs in its own change.

Build side: the sources type-check against the pinned LiteRT headers and compile under Code Analysis with `WITH_LITERT=1`; `dart analyze`, `dart format`, `clang-format` and `buildifier` are clean; the settings block parses under `protoc`; the four model checksums match the md5s of the actually-downloaded files.

### DIFF
--- a/.github/workflows/android-build-test-linux.yml
+++ b/.github/workflows/android-build-test-linux.yml
@@ -208,10 +208,6 @@ jobs:
         env:
           OFFICIAL_BUILD: false
           PERF_TEST: true
-          # TEMPORARY (revert before merge): run only stable diffusion, to
-          # validate the new LiteRT SD pipeline on one device session while
-          # the BrowserStack parallel budget is contended.
-          BENCHMARK_IDS: "stable_diffusion"
           WITH_TFLITE: 1
           WITH_LITERT: 1
           WITH_PIXEL: 1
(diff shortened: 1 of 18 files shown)


freedomtan:
@anhappdev: Is this based one @farook-edev's work? I remember he mentioned that he had work-in-progress stable diffusion work. Please sync with @farook-edev.

anhappdev:
> Is this based one @farook-edev's work? I remember he mentioned that he had work-in-progress stable diffusion work. Please sync with @farook-edev.

@freedomtan The PR is based on the last commit [added separate LiteRT backend](https://github.com/mlcommons/mobile_app_open/commit/fd09e867a84c2a6b332f5dc745a62657266b03ca) pushed to `compiled_model` by @farook-edev some days ago.

@farook-edev Do you have any new updates for LiteRT? Please push it to the `litert` branch.

In general I think for work in progress, we should open a draft pull request, so that we know who is working on what to avoid double work.

PR #1175, opened 2026-08-27
Add iOS support to mobile_back_litert (CPU + Metal), upgrade to LiteRT 2.2.0 

anhappdev:
Adds the Apple half of `mobile_back_litert`: iPhone and iPad run all nine benchmarks through the same LiteRT v2 CompiledModel code as Android, on Metal — including `stable_diffusion`, which needs the **LiteRT 2.2.0** upgrade this branch also carries.

Stacked on #1171 — merge that first. Up to date with `litert`, including #1174. Supersedes #1178.

## What it adds

- `mlperf_backend_matches_hardware` returned false off Android; it now serves a new `litert_settings_apple.pbtxt`. Nine benchmarks claimed, all selecting Metal.
- New `apple_xcframework`, minimum iOS 15.0 — LiteRT's `LITERT_MIN_IOS_VERSION`, and what the prebuilt accelerator declares. Sibling backends and the Flutter framework move up from 13.1 to match; the app already targeted 15.0.
- Metal comes from a prebuilt `libLiteRtMetalAccelerator.dylib` pinned at 2.2.0, wrapped in a framework (`ITMS-90426` rejects a bare Mach-O under `Frameworks/`; its `MinimumOSVersion` is read back out of the dylib, after a stale 14.0 drew `ITMS-90208`) and dlopened via a directory recovered with `dladdr` — iOS has no dlopen search path and `native_lib_path` is empty there.
- `Runner.entitlements` gains `kernel.increased-memory-limit` and `kernel.extended-virtual-addressing`, in a revertible commit of their own.
- A `litert` row in both iOS CI matrices; `llm-*` `quick.max_duration` 40 -> 150.
- Two fixes that are not iOS-specific: `custom_buffer_teardown.patch` regenerated with context (its offsets had made `destroy_func` unreachable, so no hardware buffer was released — host teardown 1840 -> 875 MiB), and `CreateOutputBuffers` now sizes buffers after resize, which is why `object_detection` crashed and `image_classification_v2` did not.

## Why LiteRT 2.2.0

The blocker is `stable_diffusion` on Metal. On 2.1.5 the backend emits shader source that does not compile for the diffusion model's int8 weights — `use of undeclared identifier 'q0'` — with `AllowSrcQuantizedFcConvOps` both on and off, so it is code generation rather than an option we set. The text encoder compiles there; the diffusion model does not.

It cannot be patched. The codegen ships only inside the prebuilt `libLiteRtMetalAccelerator.dylib` — `ml_drift_delegate/` in the source archive is TFLite-facing glue with no shader templates, so `patches/*.patch` has no way to reach it. Nor can the accelerator move on its own: a 2.2.0 dylib will not load into a 2.1.5 runtime, or the reverse.

So the whole pin moves, and `org_tensorflow` with it — the GPU delegate needs `absl/status:status_macros` from the Abseil that TensorFlow brings in (`WORKSPACE:186`). That coupling is what makes this a toolchain change for **every** backend and platform rather than a version bump. Patches go seven -> four, plus five for upstream gaps the old pins hid.

2.2.0 renames the bug rather than fixing it, which is the next section.

## Stable diffusion on Metal

`tools/sd_gpu/convert.py` rewrites the published v5_0 exports so the delegate accepts them (fold run-time shapes to constants, drop `BROADCAST_TO`, squeeze rank-5 tensors, lower a stale `SUB` version).

A fifth rewrite makes it *fit*. Constant tensor sharing decides whether weights are materialised or dequantised in-shader; without it the diffusion model needs **4409 MiB** against ~2885 available. Enabling it hit `error: use of undeclared identifier 'scale'` — the backend picks a scalar weight-dequant template for **per-tensor int8** weights without declaring the arguments it references. Single-op models place the fault exactly: per-tensor int8 `FULLY_CONNECTED` fails; per-axis FC and every `CONV_2D` form compile.

With the codegen out of reach, the model is the only side we control: the converter re-expresses those weights as per-axis, repeating the scale they already carry. No weight byte moves, CPU output bit-identical. Diffusion drops to **1913 MiB**. Outputs publish as `*_litert_v2.tflite`, leaving the originals for the base branch and #1174.

## Memory

iOS kills a process past a per-process cap (3376 MB on an 8 GB device) and `EXC_RESOURCE` cannot be caught. Apple-only; Android untouched.

- Sharing is on. Off, `llm-1b` needs 4497 MiB and diffusion 4409; on, 2055 and 1913.
- **Releasing a GPU-compiled model does not return its memory** — the weights are Metal buffers libmalloc never owned. On device, releasing two of three models recovered 69 MiB of ~1268, and rebuilding one stacked a second copy and crossed the limit. `set_phase` is therefore CPU-only; on Metal the three compile once and stay resident (1973 MiB of 3013, peak 2588), which also took a 20-step image from 45.5 s to 18.8 s on a macOS host.
- `llm-3b`/`llm-8b` are never offered: `llm-1b` alone costs ~2072 MiB resident from a 1229 MiB file.

## Results

Nine benchmarks in one process, no crash or relaunch. The 16 Pro is the CI run at `b2f9f166`; the two 17 Pros are a BrowserStack run of the same test package at `3a0ec360`. QPS, except `llm-*` which is tok/s.

| benchmark | 16 Pro | 17 Pro | 17 Pro Max | |
| --- | --- | --- | --- | --- |
| image_classification_v2 | 93.7 | 105.9 | 102.7 | valid |
| object_detection | 247.5 | 285.5 | 268.9 | valid |
| image_segmentation_v2 | 145.2 | 194.4 | 186.6 | valid |
| natural_language_processing | 50.7 | 61.9 | 60.4 | valid |
| super_resolution | 13.5 | 16.0 | 16.2 | valid |
| image_classification_offline_v2 | 99.0 | 140.3 | 137.9 | valid |
| stable_diffusion | 0.0133 | 0.0280 | 0.0274 | `isMinQueryMet: false` |
| llm-1b | 25.0 | 38.1 | 37.1 | `isEarlyStoppingMet: false` |
| llm-1b-instruct | 25.0 | 38.5 | 33.5 | `isEarlyStoppingMet: false` |

Every benchmark selects Metal on all three, with no CPU fallback and no memory kill. `stable_diffusion` puts 2732 of 2742 diffusion ops on the GPU: ~75 s/image on the 16 Pro and ~36 s on the 17 Pro, against ~98 s on CPU. Metal `llm-1b` measured 23.5–26.1 tok/s on the 16 Pro, against 9.27 on CPU and 3.46 on the Pixel 10 Pro GPU, TinyMMLU 41% vs 42%. The same BrowserStack build passed on the iPhone 17, 17e and Air too. iPad mini: Metal wins all six valid vision/NLP benchmarks, by 1.6x to 3.0x.

## Known limits

- The converter's equivalence check is one seeded input set per model through the CPU `Interpreter` — it does not exercise the fp16 Metal path, and is not a proof for all inputs.
- `stable_diffusion` and `llm-*` are not `isResultValid` (`isMinQueryMet` / `isEarlyStoppingMet` false; `llm-*` would need ~64 queries, ~11 min each). SD already carries an exemption in `performance_result_validity.dart`; extending it to `llm-*` is a scoring-policy call left for a reviewer.
- `llm-1b*` on the iPad mini is unmeasured on Metal, and claimed on every iOS device regardless — treat a red iPad run as expected. A tested-device allowlist is the intended fix.
- No iOS device job asserts throughput, for any backend: BrowserStack reports `iPhone17,1` while `expected_throughput.dart` is keyed on `iPhone16,2`. Adding it would make inert assertions live for `tflite` and `apple` too — a separate change.
- No CoreML or ANE path: LiteRT v2 does not expose one. CoreML stays `mobile_back_apple`'s job.


freedomtan:
@anhappdev do we have 
```
com.apple.developer.kernel.increased-memory-limit
com.apple.developer.kernel.extended-virtual-addressing
```
entitlements added?

freedomtan:
@anhappdev [Google AI Edge Gallery](https://apps.apple.com/us/app/google-ai-edge-gallery/id6749645337 Google AI Edge Gallery), which uses LiteRT-LM, could run Gemma E2B with Metal Accelerator, on iPhone 16 Pro (8 GiB DRAM) and iPhone 14 Pro (6 GiB DRAM). We should be able to run a 1B model.

freedomtan:
When I tried to build and run the app on iPhone 16 Pro, I ran into several issues. I'll send some diffs later.

anhappdev:
> @anhappdev do we have
> 
> ```
> com.apple.developer.kernel.increased-memory-limit
> com.apple.developer.kernel.extended-virtual-addressing
> ```
> 
> entitlements added?

No, neither is present.
We have those:

- aps-environment
- com.apple.developer.devicecheck.appattest-environment
- com.apple.security.app-sandbox

freedomtan:
1. I tried to run it with TestFlight <-- seems to be fine.
2. I also try to build this on macOS 27 beta and Xcode 27.x beta <-- this doesn't work out of box. I'll send diffs/patches.
3. Both LLM and Stable Diffusion are on LiteRT CPU now. <-- let's try to add memory related entitlements to if we can enable running with GPU accelerator <-- @anhappdev

anhappdev:
Status: LLM works with Metal on iOS now. SD requires model reconversion and upgrading LiteRT from v2.1.5 to v2.2.0.

anhappdev:
@freedomtan We now have both LLM and SD running with Metal. You can test it using TestFlight build 270.

freedomtan:
> @freedomtan We now have both LLM and SD running with Metal. You can test it using TestFlight build 270.

@anhappdev I tested on an iPhone 17 Pro.

two issues:
1. LLM 1B (instruct) always failed to finish (we met similar problems before, right? @farook-edev)
2. UI selection issue: when I select LLM only, sometimes what ran are two Image Classification v2 benchmarks. 

one concern:
please check with @farook-edev to see if upgrading LiteRT version is the "right" approach.

freedomtan:
To build with Xcode 27.0 on macOS 27.0. One-line change is needed.

```diff
From fed5f0bd59abbe8d93472cf6b9ce8b2e6c56faaa Mon Sep 17 00:00:00 2001
From: Koan-Sin Tan <koansin.tan@gmail.com>
Date: Mon, 14 Sep 2026 10:53:03 +0800
Subject: [PATCH] set swift version to be 6

With Xcode 27 on macOS 27, to build the coreml backend with
`make WITH_APPLE=1`, we need set the swift version. Either
`copts = ["-swift-version", "5"],`, or  `copts = ["-swift-version", "6"],` is fine.
---
 mobile_back_apple/cpp/backend_coreml/BUILD | 1 +
 1 file changed, 1 insertion(+)

diff --git a/mobile_back_apple/cpp/backend_coreml/BUILD b/mobile_back_apple/cpp/backend_coreml/BUILD
index 271a42d9..57030d3c 100644
--- a/mobile_back_apple/cpp/backend_coreml/BUILD
+++ b/mobile_back_apple/cpp/backend_coreml/BUILD
@@ -56,6 +56,7 @@ swift_library(
     srcs = [
         "coreml_util.swift",
     ],
+    copts = ["-swift-version", "6"],
     generates_header = True,
     deps = [
         ":apple_frameworks",
-- 
2.54.0 (Apple Git-157)

```

anhappdev:
@freedomtan 

> To build with Xcode 27.0 on macOS 27.0. One-line change is needed.

Thanks for the fix with Xcode 27.0 on macOS 27.0. 


> LLM 1B (instruct) always failed to finish (we met similar problems before, right? @farook-edev)

I run a test with 17 Pro on BrowserStack and it finished (see updated PR body). What is the error you got?


> UI selection issue: when I select LLM only, sometimes what ran are two Image Classification v2 benchmarks.

Yes, I noted that bug too. I can fix it in another PR.


> please check with @farook-edev to see if upgrading LiteRT version is the "right" approach.

I need to upgrade LiteRT to v2.2.0 to make SD run on Metal. 

@farook-edev Do you have any concern with upgrading LiteRT to v2.2.0?

farook-edev:
> LLM 1B (instruct) always failed to finish (we met similar problems before, right? @farook-edev)

@freedomtan That was a flutter issue, it was caused by the flutter side reading the wrong data and -as a result- canceling the instruct benchmark

> @farook-edev Do you have any concern with upgrading LiteRT to v2.2.0?

@anhappdev I would generally prefer to stick with `2.1.x` unless **absolutely** necessary, my preference is that since we're still in the experimental phase, it's best to stick to one version to avoid unexpected breakage due to API changes between 2.1 and 2.2. So please avoid dependency upgrades unless it's the only solution (no patches or a couple line edits from our side can fix it).

freedomtan:
@freedomtan to capture the log, it seems to be un-related to our previous one what crashed in Submission run.

freedomtan:
@farook-edev let's check we have the same issue to run SD on Android for the GPU accelerator on Android.

freedomtan:
for iOS 27, one more patch needed. Previous one only makes it compile. There are runtime errors. It seems, on iOS 27, UIKitCore evaluates runtime issues and raises a debugger trap(brk #0) if the app has not adopted the [UIScene lifecycle](https://developer.apple.com/documentation/uikit/transitioning-to-the-uikit-scene-based-life-cycle?language=objc). When launched in debug mode with LLDB attached, the debugger caught this breakpoint and
paused the app process.

1. Updated AppDelegate.swift to conform to FlutterImplicitEngineDelegate and implement didInitializeImplicitFlutterEngine.
2. Added UIApplicationSceneManifest with FlutterSceneDelegate to both
   Info-Debug.plist and Info-Release.plist. 

```diff
diff --git a/flutter/ios/Runner/AppDelegate.swift b/flutter/ios/Runner/AppDelegate.swift
index 62666446..c30b367e 100644
--- a/flutter/ios/Runner/AppDelegate.swift
+++ b/flutter/ios/Runner/AppDelegate.swift
@@ -2,12 +2,15 @@ import Flutter
 import UIKit
 
 @main
-@objc class AppDelegate: FlutterAppDelegate {
+@objc class AppDelegate: FlutterAppDelegate, FlutterImplicitEngineDelegate {
   override func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
-    GeneratedPluginRegistrant.register(with: self)
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
   }
+
+  func didInitializeImplicitFlutterEngine(_ engineBridge: FlutterImplicitEngineBridge) {
+    GeneratedPluginRegistrant.register(with: engineBridge.pluginRegistry)
+  }
 }
diff --git a/flutter/ios/Runner/Info-Debug.plist b/flutter/ios/Runner/Info-Debug.plist
index 2f766b88..9757f4ed 100644
--- a/flutter/ios/Runner/Info-Debug.plist
+++ b/flutter/ios/Runner/Info-Debug.plist
@@ -43,6 +43,27 @@
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>
 	<string>Main</string>
+	<key>UIApplicationSceneManifest</key>
+	<dict>
+		<key>UIApplicationSupportsMultipleScenes</key>
+		<false/>
+		<key>UISceneConfigurations</key>
+		<dict>
+			<key>UIWindowSceneSessionRoleApplication</key>
+			<array>
+				<dict>
+					<key>UISceneClassName</key>
+					<string>UIWindowScene</string>
+					<key>UISceneDelegateClassName</key>
+					<string>FlutterSceneDelegate</string>
+					<key>UISceneConfigurationName</key>
+					<string>flutter</string>
+					<key>UISceneStoryboardFile</key>
+					<string>Main</string>
+				</dict>
+			</array>
+		</dict>
+	</dict>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
diff --git a/flutter/ios/Runner/Info-Release.plist b/flutter/ios/Runner/Info-Release.plist
index 30209627..8588fc2c 100644
--- a/flutter/ios/Runner/Info-Release.plist
+++ b/flutter/ios/Runner/Info-Release.plist
@@ -39,6 +39,27 @@
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>
 	<string>Main</string>
+	<key>UIApplicationSceneManifest</key>
+	<dict>
+		<key>UIApplicationSupportsMultipleScenes</key>
+		<false/>
+		<key>UISceneConfigurations</key>
+		<dict>
+			<key>UIWindowSceneSessionRoleApplication</key>
+			<array>
+				<dict>
+					<key>UISceneClassName</key>
+					<string>UIWindowScene</string>
+					<key>UISceneDelegateClassName</key>
+					<string>FlutterSceneDelegate</string>
+					<key>UISceneConfigurationName</key>
+					<string>flutter</string>
+					<key>UISceneStoryboardFile</key>
+					<string>Main</string>
+				</dict>
+			</array>
+		</dict>
+	</dict>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>

```

freedomtan:
for the error on iOS 27, we can see LLM 1B (Instruct) finished running. It's
```
DartWorker: EXC_BAD_ACCESS (code=1, address=0xffffffffffffffff)
```
when  it tried to run compute accuracy, which is NOT supposed to do for QUICK mode.

```text
...
[mem] llm: prefill Run done: 12703 MiB left before the iOS limit
================================================
MLPerf Results Summary
================================================
SUT name : LiteRT
Scenario : SingleStream
Mode     : PerformanceOnly
90.0th percentile latency (ns) : 21143676291
90.0th first token percentile latency (ns) : 2310982875
Result is : INVALID
  Min duration satisfied : Yes
 Min queries satisfied : Yes
  Early stopping satisfied: NO
Recommendations:
TTFT Early Stopping Result:

TPOT Early Stopping Result:
 * Only processed 10 queries.
 * Need to process at least 64 queries for early stopping.

================================================
Additional Stats
================================================
QPS w/ loadgen overhead         : 0.10
QPS w/o loadgen overhead        : 0.11

Min latency (ns)                : 2584885958
Max latency (ns)                : 21143676291
Mean latency (ns)               : 8896971487
50.00 percentile latency (ns)   : 8009889708
90.00 percentile latency (ns)   : 21143676291
95.00 percentile latency (ns)   : 21143676291
97.00 percentile latency (ns)   : 21143676291
99.00 percentile latency (ns)   : 21143676291
99.90 percentile latency (ns)   : 21143676291

TPS w/ loadgen overhead         : 31.31
TPS w/o loadgen overhead        : 30.40
Min First Token latency (ns)                : 2223870708
Max First Token latency (ns)                : 2310982875
Mean First Token latency (ns)               : 2249673870
50.00 percentile first token latency (ns)   : 2237844791
90.00 percentile first token latency (ns)   : 2310982875
95.00 percentile first token latency (ns)   : 2310982875
97.00 percentile first token latency (ns)   : 2310982875
99.00 percentile first token latency (ns)   : 2310982875
99.90 percentile first token latency (ns)   : 2310982875

Min Time per Output Token (ns)                : 24243502
Max Time per Output Token (ns)                : 25399394
Mean Time per Output Token (ns)               : 24815791
50.00 percentile time to output token (ns)   : 24951047
90.00 percentile time to output token (ns)   : 25399394
95.00 percentile time to output token (ns)   : 25399394
97.00 percentile time to output token (ns)   : 25399394
99.00 percentile time to output token (ns)   : 25399394
99.90 percentile time to output token (ns)   : 25399394

================================================
Test Parameters Used
================================================
samples_per_query : 1
target_qps : 1000
ttft_latency (ns): 100000000
tpot_latency (ns): 100000000
max_async_queries : 1
min_duration (ms): 10000
max_duration (ms): 150000
min_query_count : 10
max_query_count : 0
qsl_rng_seed : 2465351861681999779
sample_index_rng_seed : 14276810075590677512
schedule_rng_seed : 3936089224930324775
accuracy_log_rng_seed : 0
accuracy_log_probability : 0
accuracy_log_sampling_target : 0
print_timestamps : 0
performance_issue_unique : 0
performance_issue_same : 0
performance_issue_same_index : 0
performance_sample_count : 33

No warnings encountered during test.

1 ERROR encountered. See detailed log.
I0000 00:00:1789545725.720184 6681055 dart_run_benchmark.cc:158] li:flutter/cpp/flutter/dart_run_benchmark.cc:158@dart_ffi_run_benchma
```
after [dart_run_benchmark.cc:158](https://github.com/mlcommons/mobile_app_open/blob/3a0ec36020792be138651af829fc65710c9cabeb/flutter/cpp/flutter/dart_run_benchmark.cc#L158)  is to LOG information.

image

image


```
(lldb) bt
* thread #31, name = 'DartWorker', stop reason = EXC_BAD_ACCESS (code=1, address=0xffffffffffffffff)
   * frame #0: 0x0000000107c00d00 backend_bridge_fw`mlperf::mobile::ifeval::remove_font_modifiers(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 176
     frame #1: 0x0000000107c0092c backend_bridge_fw`mlperf::mobile::ifeval::transform_response(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 216
     frame #2: 0x0000000107bfbb4c backend_bridge_fw`mlperf::mobile::ifeval::Instruction::IsFollowed(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, bool) const + 40
     frame #3: 0x0000000107bfb2ec backend_bridge_fw`mlperf::mobile::IFEval::ComputeSampleAccuracy(int, mlperf::mobile::ifeval::Accuracy&) + 164
     frame #4: 0x0000000107bfbf54 backend_bridge_fw`mlperf::mobile::IFEval::ComputeAccuracy() + 128
     frame #5: 0x00000001081b0160 backend_bridge_fw`dart_ffi_run_benchmark + 3404
     frame #6: 0x000000011ad07fe8 
     frame #7: 0x000000013410f20c 
     frame #8: 0x0000000134100744 
     frame #9: 0x000000012f9a60b4 
     frame #10: 0x000000012f53296c 
     frame #11: 0x000000013917649c 
     frame #12: 0x0000000139177570 
     frame #13: 0x0000000139174ef0 
     frame #14: 0x000000012f9a7e18 
     frame #15: 0x0000000130d7fbf8 
     frame #16: 0x0000000130d7f800 
     frame #17: 0x000000013912c57c 
     frame #18: 0x000000012f43ea54 
     frame #19: 0x000000012f43e5e0 
     frame #20: 0x000000013912c57c 
     frame #21: 0x000000013913253c 
     frame #22: 0x0000000139178914 
     frame #23: 0x000000012f9b15e8 
     frame #24: 0x000000012f9b0b10 
     frame #25: 0x000000012f9afef8 
     frame #26: 0x000000012f9afc10 
     frame #27: 0x000000013917dbe8 
     frame #28: 0x0000000139178dac 
     frame #29: 0x000000011ad03a84 
     frame #30: 0x000000010b9eb104 Flutter`dart::DartEntry::InvokeFunction(dart::Function const&, dart::Array const&, dart::Array const&) + 596
     frame #31: 0x000000010ba0cb78 Flutter`dart::IsolateMessageHandler::HandleMessage(std::__fl::unique_ptr<dart::Message, std::__fl::default_delete<dart::Message>>) + 1524
     frame #32: 0x000000010ba3c998 Flutter`dart::MessageHandler::HandleMessages(dart::MonitorLocker*, bool, bool) + 328
     frame #33: 0x000000010ba3c670 Flutter`dart::MessageHandlerTask::Run() + 572
     frame #34: 0x000000010bb53f08 Flutter`dart::ThreadPool::Worker::Main(unsigned long) + 412
     frame #35: 0x000000010badd6c4 Flutter`dart::ThreadStart(void*) + 272
     frame #36: 0x000000018569dd00 libsystem_pthread.dylib`_pthread_start + 136
```

PR #1178, opened 2026-09-09
Move LiteRT to 2.2.0 so stable_diffusion runs on Metal

anhappdev:
Moves the LiteRT pin to 2.2.0, which is what lets `stable_diffusion` run on the Metal delegate. Stacked on #1175 — merge that first.

On a macOS host, one 20-step image goes from 43.84 s on CPU to 18.8-19.4 s on Metal (two runs), with all three models compiled on the delegate and no CPU retry.

**Why the version has to move**

LiteRT 2.1.5 cannot compile the GPU-ready diffusion model at all — its Metal backend emits invalid shader source for the int8 weights (`use of undeclared identifier 'q0'`), with `AllowSrcQuantizedFcConvOps` on and off, so it is a code-generation bug rather than an option. The accelerator cannot be upgraded on its own either: a 2.2.0 dylib will not load into a 2.1.5 runtime, nor the reverse.

`org_tensorflow` moves with it, to `bcdab1a6` — the commit LiteRT 2.2.0 itself pins. LiteRT's GPU delegate needs `@com_google_absl//absl/status:status_macros`, which only exists in the Abseil that this TensorFlow's XLA brings in. All four TensorFlow patches apply to it unchanged.

**Patches: seven to four**

* `litert_coreml_repeatedfield_resize.patch` deleted — 2.2.0 already calls `RepeatedField::Resize` and keeps lowercase `resize()` on the float16 `std::string`, which is exactly what the patch did.
* `png-with-number-of-channels-detected` and `use_unsigned_char` folded into a regenerated `enable-png-...` patch. 2.2.0 rewrote `image_preprocessing_stage.cc` (loaders return `TfLiteStatus` and take `absl::string_view`, `ImageData::data` is a `std::vector<float>`, `ABSL_LOG`), so they had to be rewritten anyway and were only separate because they were authored months apart. Two fixes while rewriting: the decode buffer was a bare `new[]` that was never freed, and a decode failure aborted through `CHECK` rather than returning `kTfLiteError`. A hunk repointing a unit test at `grace_hopper.png` is dropped — no such file exists in the tree.
* `tensorflow-fix-file-opening-mode-for-Windows.patch` regenerated; still needed, 2.2.0 still opens the JPEG in text mode.
* `custom_buffer_teardown.patch` and `litert-internal-visibility.diff` unchanged.

**Stable diffusion**

The Metal choice names its own models — the GPU-ready exports from `tools/sd_gpu/convert.py`, published alongside `llama_q8_ekv3072_litert.tflite` — with per-delegate `*_filename` overrides, the same shape the `llm-*` Metal choices already use. The CPU choice keeps the published v5_0 files.

The pipeline also stops asking for constant tensor sharing. With sharing on, 2.2.0's Metal backend still generates invalid shader source (`use of undeclared identifier 'scale'`); bisecting the GPU options one at a time shows sharing alone is the trigger. `AllowSrcQuantizedFcConvOps` goes with it because `litert_gpu_options.h` says sharing must be true to use it, and it is not needed — the rewritten models are fully delegated without it, and dropping it also stops quantizing inputs to 8 bit. The `llm-*` pipeline keeps sharing.

**What is not proven**

* Everything above is measured on a macOS host. No rewritten model has run on an iPhone.
* Turning sharing off is not free: the header says it reduces constant-tensor allocation "even though tensors are not shared with other subgraphs", so this gives up their mmap/madvise handling on the pipeline closest to the iOS per-process limit. **The iOS memory cost is unmeasured.** It is accepted because the alternative is no Metal run at all.
* `org_tensorflow` moving changes the toolchain for **every** backend and platform, not just iOS. Only the litert backend was built and run here, so the full CI matrix is the real check.
* The converter's equivalence check runs one seeded input set per model through the CPU `Interpreter`. It does not exercise the fp16 Metal path these models are selected for.
* `llm-1b` still runs on Metal with no fallback, but its throughput was not remeasured against 2.1.5.

A Codex review of this branch caught two defects, both fixed here: an unversioned accelerator cache path that would have paired a stale 2.1.5 dylib with the 2.2.0 runtime on any machine with a warm `output/`, and the original claim that a single signature made dropping sharing cheap.


anhappdev:
Superseded: the 2.2.0 work is being fast-forwarded into litert-ios directly rather than landing as a separate PR.

PR #1179, opened 2026-09-15
fix: bump protobuf runtime to 6.1.0 to match protoc_plugin 25.1.0

anhappdev:
Master's unit-test job broke without a code change: `protoc_plugin` 25.1.0 was published on 2026-09-11 and every build entry point activates the plugin with a floating `^25.0.0`, so CI now generates protos with 25.1.0 while `pubspec.lock` still pinned the runtime to `protobuf` 6.0.0.

25.1.0 emits `static $pb.GeneratedMessage $_createMessage() => X._();` and passes it to `$_defaultFor`. That signature changed between runtimes — 6.0.0 takes `T Function()`, 6.1.0 takes `GeneratedMessage Function()` — so the generated code no longer type-checks against 6.0.0. 43 analyzer errors, all six unit-test files fail to load.

Verified locally, all four combinations of generator × runtime:

| protoc_plugin | protobuf | result |
| --- | --- | --- |
| 25.0.0 | 6.0.0 | clean (what CI used to do) |
| 25.1.0 | 6.0.0 | 43 errors — this failure |
| 25.1.0 | 6.1.0 | clean |
| 25.0.0 | 6.1.0 | clean |

So bumping the runtime rather than pinning the generator is the fix that widens compatibility: 6.1.0 accepts code from both plugin versions, so anyone with 25.0.0 still globally activated keeps working. Both packages need only Dart SDK `^3.7.0`, well under CI's Dart 3.12. Full unit suite passes (70/70).

One caveat worth knowing: this can recur. The generator floats (`^25.0.0`) while the runtime is pinned by the lockfile, so the next plugin release that requires a newer runtime breaks master the same way. Pinning the activation to the exact locked version in the workflows, Dockerfiles, `ci_post_clone.sh` and the env-setup docs would close that off, at the cost of a coupled bump each time.


freedomtan (approved):
LGTM

